PR #115, opened 2015-04-03
Getting feet wet, added origin host header for sites which require it ma...

hammer2j2:
Getting feet wet, added origin host header for sites which require it match the site name.

I changed this because if routing to other sites, for example, example.org, or google.com, there will be a forbidden response due to seeing a host header of localhost:8080.  Below is the tcpdump to strings trace:

google
GET /?q=edge+router HTTP/1.1
Host: localhost:8080
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:24.0) Gecko/20100101 Firefox/24.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Connection: keep-alive
HTTP/1.1 403 Forbidden
Connection: close
Via: HTTP/1.1 proxy1732

### DIFF
--- a/zuul-simple-webapp/src/main/groovy/filters/pre/PreDecoration.groovy
+++ b/zuul-simple-webapp/src/main/groovy/filters/pre/PreDecoration.groovy
@@ -44,6 +44,9 @@ class PreDecorationFilter extends ZuulFilter {
         // sets origin
         ctx.setRouteHost(new URL("http://apache.org/"));
 
+        // set origin host header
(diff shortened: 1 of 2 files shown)


NiteshKant:
@mikeycohen @kerumai this change looks ok to me and it is only in the sample app. Are you guys ok with merging this?

NiteshKant:
Thanks @hammer2j2

hammer2j2:
Thanks for the merge @NiteshKant

PR #134, opened 2015-07-16
Check isFilterDisabled in runFilter.

ghenkes:
Noticed runFilter in ZuulFilter doesn't call the function isFilterDisabled (as stated in the javadoc) and calls the Archaius property directly. This means that if the function is overridden in a filter extending ZuulFilter it is currently being ignored.


NiteshKant:
@mikeycohen @kerumai is this ok to merge? Makes sense to me.

NiteshKant:
Thanks @ghenkes

PR #137, opened 2015-07-31
Report ZuulExceptions through logger in ZuulServletFilter

ghenkes:
In ZuulServletFilter ZuulExceptions are just getting printed rather than getting added to the log like in ZuulServlet.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/ZuulServletFilter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/ZuulServletFilter.java
@@ -26,8 +26,6 @@ import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 import org.mockito.runners.MockitoJUnitRunner;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import javax.servlet.*;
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/filters/ZuulServletFilter.java
```diff
@@ -109,7 +113,7 @@ void init(HttpServletRequest servletRequest, HttpServletResponse servletResponse
     void error(ZuulException e) {
         RequestContext.getCurrentContext().setThrowable(e);
         zuulRunner.error();
-        e.printStackTrace();
+        LOG.error(e.getMessage(), e);
```
spencergibb: My preference would be to remove `e.printStackTrace()` altogether and let the 'error' filters determine if they want to log the error.

NiteshKant: +1

ghenkes:
@spencergibb @NiteshKant removed logging the exception in ZuulServletFilter. Also removed similar line in ZuulServlet. Added logging to the example ErrorResponse filter in zuul-netfilx-webapp.

NiteshKant:
Thanks @ghenkes !

PR #142, opened 2015-08-05
Fixed Issue 141 - Request Parameter with empty value being stripped

kcbaltz:
Fixed an issue that caused request parameters with no value (e.g. ?wsdl
or ?wsdl=) to not be included in the result of getQueryParams()


spencergibb:
:+1:

mikeycohen:
Looks good to me. Thanks!

PR #144, opened 2015-08-18
replace header with zuul request header, fix multiple header bug

tootedom:
Hi there,  

I noticed a couple of issues with the `buildZuulRequestHeaders` method:

- The replacement of request headers with the headers from getZuulRequestHeaders doesn't work as the header in the Header array isn't in LowerCase (https://github.com/Netflix/zuul/blob/1.x/zuul-netflix-webapp/src/main/groovy/filters/route/ZuulHostRequest.groovy#L336), so the find doesn't match any headers.  I noticed this whist trying to reverseProxy and change the Host header via `RequestContext.getCurrentContext().setRouteHost(..)`

- The find on an ArrayList isn't performant, as it will just be a linear search of the ArrayList for the matching header.

- The building of the headers doesn't taking into consideration multiple headers in the request header.

Hope that makes sense
cheers
/dom

### DIFF
--- a/zuul-netflix-webapp/src/main/groovy/filters/route/ZuulHostRequest.groovy
+++ b/zuul-netflix-webapp/src/main/groovy/filters/route/ZuulHostRequest.groovy
@@ -319,34 +319,36 @@ class ZuulHostRequest extends ZuulFilter {
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


mikeycohen:
Looks good. Thanks!

PR #161, opened 2015-09-18
Adding getSanitizedQueryString()

jusroberts:
Hi all,

This PR cleans up the request query to allow an HttpClient to process query strings that come in with invalid characters ({, }, |, ^, etc) without failing on a `org.apache.http.client.ClientProtocolException`. 

Before this code change, if you try to hit `localhost:8080/?a={}`, you will receive a blank page caused by the above exception. This is poor behavior, as most web servers are able to handle these invalid characters without rendering an error. To fix this, we re-encode the query string so that it will have the proper characters according to https://tools.ietf.org/html/rfc3986#page-12 (Sections 2.1-2.3).

```java
String getSanitizedQueryString() {
    String encoding = "UTF-8"
    String decodedQueryString = URLDecoder.decode(getQueryString(), encoding)
    String query = new URI(null, null, null, decodedQueryString, null).toString()
    if (query.equals("?"))
        query = ""
    return query
}
```

You will notice that I use `URLDecoder` to decode the query string, but then use `URI` (and not `URLEncoder`) to re-encode it. This is done to follow best practices outlined on StackOverflow [[1]](http://stackoverflow.com/questions/10786042/java-url-encoding-of-query-string-parameters) [[2]](http://stackoverflow.com/questions/14321873/java-url-encoding-urlencoder-vs-uri).

Questions, comments, concerns?


NiteshKant:
@jusroberts thanks for the PR, your code is outdated, you need to rebase to the master.

jusroberts:
Fixed. I seemed to have forked an old copy.

jusroberts:
I combined `getSanitizedQueryString()` with `getQueryString()` because `getQueryString()` isn't used by anything other than the forward() method.

NiteshKant:
Thanks @jusroberts

PR #168, opened 2015-10-05
Support RibbonCommand group key parameter

mattnelson:
This change was discussed on the HystrixOSS google group.
https://groups.google.com/forum/#!searchin/hystrixoss/zuul/hystrixoss/LM8zuXSYW-w/HOIjaXWfhmsJ

As is all RibbonCommands are in the same group, which means they use the same circuit and are reported in a single bucket to the hystrix dashboard. I added another constructor and remained passive with the existing functionality by using the simple class name when not supplied.

### DIFF
--- a/zuul-netflix/src/main/java/com/netflix/zuul/dependency/ribbon/hystrix/RibbonCommand.java
+++ b/zuul-netflix/src/main/java/com/netflix/zuul/dependency/ribbon/hystrix/RibbonCommand.java
@@ -10,8 +10,15 @@ import com.netflix.hystrix.HystrixCommandProperties;
 import com.netflix.niws.client.http.RestClient;
 import com.netflix.zuul.constants.ZuulConstants;
 import com.netflix.zuul.context.NFRequestContext;
+import com.netflix.zuul.dependency.httpclient.hystrix.HostCommand;
 
 import javax.ws.rs.core.MultivaluedMap;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


mattnelson:
Looks like the same change should probably be applied to HostCommand as well.

https://github.com/Netflix/zuul/blob/zuul-1.0.28/zuul-netflix/src/main/java/com/netflix/zuul/dependency/httpclient/hystrix/HostCommand.java#L34

NiteshKant:
Thanks @mattnelson. Do you intend to make changes to `HostCommand` as well? I think it makes sense.

mattnelson:
Updated HostCommand and added unit tests to validate the changes and that the current constructors are still passive.

zuul-netflix/src/main/java/com/netflix/zuul/dependency/ribbon/hystrix/RibbonCommand.java
```diff
@@ -53,13 +61,27 @@ public RibbonCommand(String commandKey,
...
+                    MultivaluedMap<String, String> params,
+                    InputStream requestEntity) throws URISyntaxException {
+        
+        super(Setter.withGroupKey(HystrixCommandGroupKey.Factory.asKey(groupKey))
+                    .andCommandKey(HystrixCommandKey.Factory.asKey(commandKey)).andCommandPropertiesDefaults(
+           // we want to default to semaphore-isolation since this wraps
+           // 2 others commands that are already thread isolated
+           HystrixCommandProperties.Setter().withExecutionIsolationStrategy(HystrixCommandProperties.ExecutionIsolationStrategy.SEMAPHORE)
+                   .withExecutionIsolationSemaphoreMaxConcurrentRequests(DynamicPropertyFactory.getInstance().
+                           getIntProperty(ZuulConstants.ZUUL_EUREKA + commandKey + ".semaphore.maxSemaphores", 100).get())));
```
mattnelson: I wasn't sure what to do about the semaphore properties, it seems like this should be using the group key.

mattnelson:
@NiteshKant Have you had a chance to review the added changes?

zuul-netflix/src/main/java/com/netflix/zuul/dependency/ribbon/hystrix/RibbonCommand.java
NiteshKant: I actually am not sure why do we need to even specify these values in code and not follow the hystrix hierarchical configuration pattern. So, a command would inherit the properties from the group if they don't exist at the command level.

The reason for not using the group here explicitly would be is that different commands in the same group may have different semaphore limits. I would leave it as is.

mattnelson: I'm good with that.

NiteshKant:
Thanks @mattnelson for the PR!

Regd: release of this fix, I am struggling with the release of 1.x branch as few tests fail on the CI server, will try it again.

Issue #169, opened 2015-10-06
setOriginContentLength in RequestContext fails for all files larger than 2.15 GB


chenopodium:
setOriginContentLength in RequestContext uses Integer.parseInt to parse the content length. However, this will fail for any file larger than 2.15GB (max int is 2147483647).
https://github.com/Netflix/zuul/blob/1.x/zuul-core/src/main/java/com/netflix/zuul/context/RequestContext.java

NiteshKant:
@chenopodium you are correct, do you want to send a PR?

chenopodium:
I just did ;-) Best wishes, Chantal


On Tue, Oct 6, 2015, at 06:53 PM, Nitesh Kant wrote:
> @chenopodium[1] you are correct, do you want to send a PR?


> — Reply to this email directly or view it on GitHub[2].


>



Links:

  1. https://github.com/chenopodium
  2. https://github.com/Netflix/zuul/issues/169#issuecomment-145925903

PR #170, opened 2015-10-06
https://github.com/Netflix/zuul/issues/169

chenopodium:
setOriginContentLength in RequestContext fails for all files larger than 2.15 GB
Changed to long


NiteshKant:
Thanks @chenopodium

PR #199, opened 2016-01-21
Handle query params without an `=`

ausmith:
If a query string contained a key pair without an `=` in the string then
it would cause an array out of bounds exception because the split was
not wrapped defensively to test for the presence of an `=` first.

Example: ?&abc=123&xyz
Split on `&` would look like: `['', 'abc=123', 'xyz']`
Both cases are now handled where the string is missing `=`

This defensive programing was unfortunately missed in my previous PR https://github.com/Netflix/zuul/pull/198

### DIFF
--- a/zuul-simple-webapp/src/main/groovy/filters/route/SimpleHostRoutingFilter.groovy
+++ b/zuul-simple-webapp/src/main/groovy/filters/route/SimpleHostRoutingFilter.groovy
@@ -263,14 +263,15 @@ class SimpleHostRoutingFilter extends ZuulFilter {
             if (rebuiltQueryString.length() > 0) {
                 rebuiltQueryString = rebuiltQueryString + "&"
             }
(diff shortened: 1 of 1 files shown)


zuul-simple-webapp/src/main/groovy/filters/route/SimpleHostRoutingFilter.groovy
```diff
@@ -263,14 +263,15 @@ class SimpleHostRoutingFilter extends ZuulFilter {
             if (rebuiltQueryString.length() > 0) {
                 rebuiltQueryString = rebuiltQueryString + "&"
             }
-            def (name,value) = keyPair.split("=", 2)
```
ausmith: Specifically, this was causing an array out of bounds exception if there was no `=` in the query param.

kerumai:
Thanks Aaron.

Issue #96, opened 2014-11-17
Launching zuul-netflix webapp fails with java.lang.ArrayIndexOutOfBoundsException


pataprogramming:
When running `zuul-netflix-webapp` with `../gradlew jettyRun` on my system (Ubuntu 14.04.1 with Oracle JDK 1.7.0_72-b14), the launch fails with an `ArrayIndexOutOfBoundsException` reproduced in full at the end of this issue report. This is apparently due to a dependency on `turbine-core` that does not fix a particular version; when this started failing on my system on Friday, I traced it to `turbine-core-2.0.0-DP.1.jar`.

Unlike the other dependencies specified in the `netflix-webapp` subdirectory (which specify a particular library version), the Turbine dependency is declared as 

```
compile 'com.netflix.turbine:turbine-core:[0.4,)'
```

Changing this to a fixed, lower version of `turbine-core` fixes this problem. (I tried 0.5 in my tests, which worked fine.)

```
> Building > :zuul-netflix-webapp:jettyRun > Starting Jetty > Resolving dependencies ':zuul-netflix-webapp:zuul-netflix-webapp:jettyRun
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/user/.gradle/wrapper/dists/gradle-1.1-bin/13d7lnhcrghv2i5e54el41jpgr/gradle-1.1/lib/logback-classic-1.0.6.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/user/.gradle/caches/artifacts-14/filestore/org.slf4j/slf4j-log4j12/1.7.2/jar/7539c264413b9b1ff9841cd00058c974b7cd1ec9/slf4j-log4j12-1.7.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/user/.gradle/caches/artifacts-14/filestore/org.slf4j/slf4j-simple/1.7.7/jar/8095d0b9f7e0a9cd79a663c740e0f8fb31d0e2c8/slf4j-simple-1.7.7.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
 WARN 16:23:53,061 No URLs will be polled as dynamic configuration sources.
[...]

Issue #23, opened 2014-05-13
Dependency version issues


jsamarziya:
We attempted to upgrade our project's dependency on zuul-core but found issues with the POM for version 1.0.21. Some of the dependency versions in the zuul-core POM appear to be invalid - versions such as `+` and `latest.release`, for example, are not accepted by maven 3.0.5 or 3.2.1.

PR #212, opened 2016-03-28
Remove unused turbine-core dependency

ghenkes:
Turbine-core is included as a compile dependency of zuul-netflix, but doesn't appear to be used.


mattnelson:
Looks like this turbine dependency has caused several other issues #23 and #96

spencergibb:
:+1:

mikeycohen:
Looks good to me. Thanks!

PR #220, opened 2016-05-06
Move test to test directory, update ribbon and hystrix libraries.

jaume-pinyol:
Hi,

I had these changes in my repo and I thought that maybe it could be interesting to try to get them merged.

I know it's a long commit but it's because I've moved all the tests to src/test/.... instead of being with the main code.

So the changes that I've done are:
* Force Java 8.
* Update hystrix, servo, archaius, and ribbon libraries.
* Update jersey libraries.
* Add an example in zuul-netflix-webapp of archaius fetching properties from dynamodb.
* Change some println lines in some filters for LOGGER.info().
* Update to gradle 2.10.
* Allow to change Cassandra cql and target version with a property.
* Add another constructor in RibbonCommand to be able to choose the commandKey. So in hystrix-dashboard multiple command can be shown separately.

I think this is all. I hope you find it useful ;)

Best Regards,
Jaume Pinyol


spencergibb:
I'm going to vote no for java8 on 1.x

mikeycohen:
@jaume-pinyol I realize that unit test code in classes is a style preference, but we do prefer having the unit tests in the code. This keeps logic and testing close. I am going to reject the pull request because of this..

PR #231, opened 2016-06-17
Fixed integer overflow in ZuulFilter.compareTo(..)

sfussenegger:
the current implementation of `ZuulFilter.compareTo(..)` is vulnerable to Integer overflows:

```java
    public int compareTo(ZuulFilter filter) {
        return this.filterOrder() - filter.filterOrder();
    }
```

In the Javadoc of `filterOrder()` is no mention of valid values, so I assume that positive and negative values including `Integer.MAX_VALUE` and `Integer.MIN_VALUE` are supported.

the simple fix is to use `Integer.compare(int,int)` instead:

```java
    public int compareTo(ZuulFilter filter) {
        return Integer.compare(this.filterOrder(), filter.filterOrder());
    }
```


kerumai:
Good catch @sfussenegger . Thanks!

PR #251, opened 2016-08-08
Prevent instantiation of constant utility classes

gzurowski:


mikeycohen:
I'll merge this, however I think this is pedantic.

PR #256, opened 2016-08-13
Apply field encapsulation in ZuulApplicationInfo and refactor client code accordingly

gzurowski:


mikeycohen:
In general, changing variable access is not recommended as it may have consequences outside of this project

PR #257, opened 2016-08-14
Standardize use of logger across all classes

gzurowski:
Use the same variable name and modifiers for loggers across all classes


mikeycohen:
Changing access from protected to private can have consequences outside of this project. Not sure why this would need to be done..

PR #259, opened 2016-08-14
Make class fields private

gzurowski:


mikeycohen:
Not sure making this private does anything useful. This could affect other code.

PR #260, opened 2016-08-14
Encapsulate fields of class ZuulException

gzurowski:


mikeycohen:
Encapsulation and changing access privilege will have consequences outside of this project.

PR #275, opened 2016-10-21
Preserve query parameter order

lsimons:
SoapUI mock SOAP services can have constructs like this one in their top-level WSDL:

``` xml
<wsdl:import namespace="http://example.com/" location="/example?WSDL&amp;interface=ExampleWSPortTypeBinding&amp;part=ExampleWebServiceImpl.wsdl">
    </wsdl:import>
```

where the order of the query parameters matters. This is not supported by zuul 1.2.2 and 1.4.0 whereas basically all other reverse proxies support it fine.

In general, IMHO, query parameters are an ordered list of key/value pairs.

Also put a backport for 1.2.2 (used by released spring-cloud-netflix starter atm) at https://github.com/lsimons/zuul/tree/v1.2.21 .

Patch obviously not applicable to zuul 2.x with all its reactive magic.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/util/HTTPRequestUtils.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/util/HTTPRequestUtils.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


jnystromdev:
+1  We have some legacy systems that we're trying to route to through Zuul and are noticing the same problem.

ervansetiawan:
We were going to submit an exact same merge request before we saw this one. Can this be merged, please? It will benefit us as well.

mikeycohen:
Looks good to me.

mattnelson:
Just discovered this issue as well. Doesn't this only address half of the problem? This will address the ordering of values relative to the same key, but not the overall ordering of all the keys.

The ordering of the `a` values is consistent, but the placement of `b` has changed.
```
?a=1&a=2&b=1
?a=1&b=1&a=2
```

PR #277, opened 2016-10-28
Make sure disable filter prop name is correct on init

spencergibb:
If you want to use a field for the filter type, the disabled prop name that gets set is `zuul.<filterTyle>.null.disable`. This change moves initialization to the first use of `isFilterDisabled()` using an `AtomicReference`.

The following test failed before this change:

```
        @Test
        public void testFilterTypeAsField() {
            class TestZuulFilter extends ZuulFilter {

                final String filterType;

                public TestZuulFilter(String filterType) {
                    this.filterType = filterType;
                }

                @Override
                public boolean shouldFilter() {
                    return false;
                }

                @Override
                public Object run() {
                    return null;
                }

                @Override
                public String filterType() {
                    return filterType;
                }

                @Override
                public int filterOrder() {
                    return 0;
                }
            }

            TestZuulFilter filter = new TestZuulFilter("pre");
            assertFalse(filter.isFilterDisabled());
            // assumes filterDisabled is package private
            String filterName = filter.filterDisabled.getName();
            assertEquals("zuul.TestZuulFilter.pre.disable", filterName);
        }
```

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/ZuulFilter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/ZuulFilter.java
@@ -24,9 +24,13 @@ import org.junit.Test;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


mikeycohen:
Looks fine to me.

PR #281, opened 2016-11-07
Cleanup ZuulHttpProcessor

otoolec:
Doing a little cleanup for ZuulHttpProcessor. I mainly wanted to add types to SessionContextFactory but ended up making fields final because of the existing comment:

> Ensure that this is initialized.


artgon:
No longer used in Zuul 2.1

PR #288, opened 2016-12-19
remove manageFiles() from init(), in case it throws exception.

135f2l:
manageFiles() will throw exceptions, which will make the init function crash. Let startPoller() do all things will be fine.


artgon:
Why would you want to start up if `manageFiles()` fails?

Issue #294, opened 2017-01-06
IllegalArgumentException happens in BufferingOutbound.groovy and BufferingInbound.groovy on Zuul 2.x


tmack8001:
From Zuul/2.x running `cd zuul-servletapi-sample && ../gradlew jettyRun` there are two IllegalArgumentExceptions thrown when attempting to load the inbound and outbound filters Buffering*.groovy within the FilterLoader (stack traces are below).

```
22527 [FilterFileManager_ProcessFiles-1] ERROR com.netflix.zuul.FilterLoader - Error loading filter! Continuing. file=/Users/tmack/dev/projects/platform-services/stargate-evals/zuul-2.x/zuul-servletapi-sample/src/main/filters/outbound/BufferingOutbound.groovy
java.lang.IllegalArgumentException: No enum constant com.netflix.zuul.filters.FilterType.out
	at java.lang.Enum.valueOf(Enum.java:238)
	at org.codehaus.groovy.runtime.typehandling.ShortTypeHandling.castToEnum(ShortTypeHandling.java:53)
	at outbound.BufferingOutbound.filterType(BufferingOutbound.groovy:38)
	at com.netflix.zuul.filters.BaseFilter.disablePropertyName(BaseFilter.java:67)
	at com.netflix.zuul.filters.BaseFilter.<init>(BaseFilter.java:48)
	at com.netflix.zuul.filters.MessageBodyBufferFilter.<init>(MessageBodyBufferFilter.groovy)
	at outbound.BufferingOutbound.<init>(BufferingOutbound.groovy)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at com.netflix.zuul.DefaultFilterFactory.newInstance(DefaultFilterFactory.java:34)
	at com.netflix.zuul.FilterLoader.putFilter(FilterLoader.java:149)
	at com.netflix.zuul.FilterFileManager.lambda$processGroovyFiles$0(FilterFileManager.java:186)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #295, opened 2017-01-06
correcting FilterType to valid enums

tmack8001:
This is addressing the issue reported in https://github.com/Netflix/zuul/issues/294 so that the BufferingOutbound.groovy and BufferingInbound.groovy filters return correct enum for the filterType() method.


artgon:
No longer used in 2.1, will review when building new sample app.

tmack8001:
is the intention of removing the sample app to move it into another repository outside of the "zuul-core" project?

artgon:
It needs to be rebuilt using the new framework.

PR #312, opened 2017-03-08
correct Ippon blog URL

hdurix:
Previous URL was leading nowhere


mikeycohen:
Thanks!

mikeycohen:
Feel free to send more zuul related links/articles our way. We will link them here.

PR #353, opened 2017-08-29
Add generic type and protected getters to RibbonCommand

mattnelson:
I have a use case in which I want to use RibbonCommand but I don't want to execute with the load balancer. Default implementation if left unchanged.

- Changed class signature to accept a generic to indicate the type of the client. Due to type erasure this is a passive change
- Added protected methods for member fields
- Added protected method to delegate client execution when overridden

### DIFF
--- a/zuul-netflix-webapp/src/main/groovy/filters/route/ZuulNFRequest.groovy
+++ b/zuul-netflix-webapp/src/main/groovy/filters/route/ZuulNFRequest.groovy
@@ -15,6 +15,7 @@
  */
 package filters.route
 
+import com.netflix.client.AbstractLoadBalancerAwareClient
 import com.netflix.client.ClientException
 import com.netflix.client.ClientFactory
 import com.netflix.client.IClient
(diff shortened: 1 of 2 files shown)


mattnelson:
@mikeycohen @qiangdavidliu 
Could I get a review on this?

qiangdavidliu:
Let me ping the zuul guys.

artgon:
@mattnelson adding this new generic to the `RibbonCommand` class will cause unchecked warnings everywhere it's currently used. Can you clean those up along with this change? 

Also, with the getters added, all the member fields can be marked private.

Otherwise it looks good to me.

mattnelson:
> Also, with the getters added, all the member fields can be marked private.

I don't mind making this change because I agree with it, but it would be non-passive since the class/members are not final. A consumer could've been abusing that fact with their own implementation in the `com.netflix.zuul.dependency.ribbon.hystrix` package to effectively have setters on the package scoped non-final members.

artgon:
@mattnelson one more place here: https://github.com/Netflix/zuul/blob/8ede1c3fec67df86770eaa0407b6912ad236e5c6/zuul-netflix-webapp/src/main/groovy/filters/route/ZuulNFRequest.groovy#L147 😄 

I agree that it's not backwards compatible but should only impact people who are intentionally extending that class.

mattnelson:
> one more place here:

Done in dbc311b

mattnelson:
@artgon When can we expect this to be released?

artgon:
@mattnelson I'll make a release next week

artgon:
Released: https://bintray.com/netflixoss/maven/zuul/1.3.1

PR #354, opened 2017-08-31
Support @Inject on Filters

quidryan:
Since Zuul already has a dependency on Guice, it made sense to leverage Guice to inject dependencies into Filters. This functionality is essentially the same as putting @Singleton objects into the SessionContext and pulling them out, but it's done just once on construction.

Since we're still pulling from what is in essence a global DI context, that  means the filters using @Inject are still considered stateless. This PR does NOTHING to make filters stateful. 

* Convert FilterLoader to be injectable, while maintaining the older constructor, to allow a FilterFactory to be provided
* Create GuiceFilterFactory that will leverage Guice to instantiate the filters, which will allow @Inject to work on Filters. Use Governator Unit test runner to prove that it'll work in the context of a Guide.
* Move FilterFileManagerConfig to a @Provides method, and broke apart the loading of class names and files, with tests.
* Added loading by package name (used internally) to FilterFileManagerConfig


quidryan:
Where would the accompanying documentation go?

quidryan:
@twicksell I'm running in archaius2-archaius1-bridge issues when running on the tests. I'm going to move back to System.setProperty to get the test to work. But I'd love to know how to easily migrate projects like Zuul to archaius2, without having to change the world in regards to static DynamicIntProperty fields.

quidryan:
Updating to be against 3.x

quidryan:
@raksoras Sorta. The master branch is failing with errors like:

```
/Users/jryan/Projects/stash/EDGE/edgezuul/oss/zuul-servletapi-sample/src/main/java/com/netflix/zuul/servlet/ZuulBootStrap.java:26: error: package com.netflix.zuul.ribbon does not exist
```

Which means that my branch also breaks with the same error. If I remove the samples from the build, zuul-core fails for unrelated tests. I can hold off or merge and wait for @sunnysingh85 to make a PR that introduces a main method based sample, and rips out servlet-api.

quidryan:
I'm not sure 3.x is going to build on it's own, let's merge this anyways.

PR #391, opened 2018-01-05
Disambiguate this project from others named zuul

### LINKED ISSUE #315: Namespace confusion is happening with OpenStack's Zuul project
Hi! I work on the Zuul project started by OpenStack developers back in 2012:

https://github.com/openstack-infra/zuul/tree/1.1.0

We've known for a while that there was a name collision when Netflix released its Zuul project. However, we may need to do something to disambiguate, as we have at least one example of confusion around the name:

https://thenewstack.io/ibm-openstack-engineer-urges-cncf-consider-augmenting-jenkins-zuul/

We should work together to find a solution to this. Thanks!

SpamapS:
There are at least two projects named Zuul that are not this one. Provide links in the README for anyone who has stumbled into the wrong Github.

Closes #315

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,14 @@
 Zuul is an edge service that provides dynamic routing, monitoring, resiliency, security, and more.
 Please view the wiki for usage, information, HOWTO, etc https://github.com/Netflix/zuul/wiki
 
+If you are looking for the OpenStack related test automation tool that is also named Zuul, you can find it here:
+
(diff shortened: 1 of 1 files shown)


artgon:
I think you need to give your users a bit more credit. Reading the first line of the description of both of these projects should make it abundantly clear which one is a proxy and which one is an automated testing framework.

PR #394, opened 2018-01-09
fix HTTPRequestUtils.getRequestHeaderMap

Kookey:
fix HTTPRequestUtils.getRequestHeaderMap

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/util/HTTPRequestUtils.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/util/HTTPRequestUtils.java
@@ -298,6 +298,21 @@ public class HTTPRequestUtils {
         }
 
         @Test
+        public void testgetRequestHeaderMap(){
+            when(request.getHeaderNames()).thenReturn(new StringTokenizer("accept;language",";"));
+            when(request.getHeaders("accept")).thenReturn(new StringTokenizer("1;2;3;4",";"));
+            when(request.getHeaders("language")).thenReturn(new StringTokenizer("zh;ch;ab;de;japan",";"));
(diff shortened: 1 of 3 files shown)


artgon:
This looks good, but can you please add a test to exercise the different branches of logic there.

zuul-core/src/main/java/com/netflix/zuul/util/HTTPRequestUtils.java
```diff
@@ -133,15 +133,31 @@ public String getFormValue(String sHeaderName) {
...
+                if (name != null && !name.isEmpty()){
+                    Enumeration values = request.getHeaders(name);
+                    if (values == null){
+                        continue;
+                    }
+                    if (headers.get(name) != null){
+                        while (values.hasMoreElements()){
+                            Object value = values.nextElement();
+                            if (value == null){
+                                continue;
```
artgon: Instead of continuing, you can do the inverse of the if statement and add it then. It'll make the code more readable. You can replace all of the `continue`s that way.

Kookey: yes,you are right

zuul-netflix-webapp/src/main/java/com/netflix/zuul/StartServer2.java
```diff
@@ -185,7 +185,7 @@ void initZuul() throws Exception, IllegalAccessException, InstantiationException
 
         LOG.info("Starting Groovy Filter file manager");
         final AbstractConfiguration config = ConfigurationManager.getConfigInstance();
-
+        Thread.sleep(40000);
```
carl-mastrangelo: What is this for?

PR #399, opened 2018-01-15
Addl ctor for ZuulException to support all args

quidryan:


zuul-core/src/main/java/com/netflix/zuul/exception/ZuulException.java
```diff
@@ -38,6 +38,17 @@ public ZuulException(String sMessage, Throwable throwable, String errorCause) {
         this.errorCause = errorCause;
     }
 
+    /**
+     * Source Throwable, message, status code and info about the cause
+     * @param sMessage
+     * @param throwable
+     * @param errorCause
+     */
```
artgon: missing param :)

quidryan: Would it be better if I removed all the @param's? Do they really add necessary documentation?

artgon: You can get rid of the params but describe the method better.

carl-mastrangelo:
@quidryan Are you still interested in this PR?   I am going through and trying to clean up our PRs and either submit them or close them out.

PR #401, opened 2018-01-22
Fix for cross-compile(from java 8 to java 7) in RequestContext(ConcurrentHashMap)

weslleyandrade:
This PR propose a solution to the problem in cross compilation.
When the zuul is compiled in jdk 8 and run in jdk 7 the exception it is launched:
"Caused by: java.lang.NoSuchMethodError: com.netflix.zuul.context.RequestContext.keySet()Ljava/util/concurrent/ConcurrentHashMap$KeySetView;"

The problem it is more detailed in: https://gist.github.com/AlainODea/1375759b8720a3f9f094


vspassion007:
please merge it, big issue for Java 7

carl-mastrangelo:
Is there still interest @weslleyandrade  in this PR?

PR #405, opened 2018-02-07
runtime tunable concurrency for async filters

raksoras:
A slow running async filter can cause high number of concurrent requests to be piled up on frontend. This PR introduces FP controlled concurrency limit that can be tuned for individual filters + single FP to disable filter concurrency limits altogether if we find a bug/leak in the implementation.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
@@ -171,10 +171,13 @@ public abstract class BaseZuulFilterRunner<I extends ZuulMessage, O extends Zuul
                 return (outMesg != null) ? outMesg : filter.getDefaultOutput(inMesg);
             }
 
-            //async filter
+            // async filter
+            filter.incrementConcurrency();
+            final FilterChainResumer resumer = new FilterChainResumer(inMesg, filter, snapshot, startTime);
             filter.applyAsync(inMesg)
-                    .observeOn(Schedulers.from(getChannelHandlerContext(inMesg).executor()))
-                    .subscribe(new FilterChainResumer(inMesg, filter, snapshot, startTime));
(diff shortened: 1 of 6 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -171,10 +171,13 @@ protected final O filter(final ZuulFilter<I, O> filter, final I inMesg) {
...
 
-            //async filter
+            // async filter
+            filter.incrementConcurrency();
+            final FilterChainResumer resumer = new FilterChainResumer(inMesg, filter, snapshot, startTime);
             filter.applyAsync(inMesg)
-                    .observeOn(Schedulers.from(getChannelHandlerContext(inMesg).executor()))
-                    .subscribe(new FilterChainResumer(inMesg, filter, snapshot, startTime));
+                .observeOn(Schedulers.from(getChannelHandlerContext(inMesg).executor()))
+                .doOnUnsubscribe(resumer::decrementConcurrency)
```
artgon: To clarify, this unsubscribe will be called after this individual async filter completes/errors, correct?

raksoras: "If" it is called. I don't think we actually unsubscribe the observer in normal case. This is paranoid safety measure so that no concurrency count may leak. Note that it will be safe for it (decrementConcurrency) to be called multiple times as FilterChainResumer internally keeps state of decrement operation and does it only once on first decrement.

artgon: I was more concerned about this counter being incremented/decremented per filter instead of the entire filter chain.

raksoras: Ah, yes, the FilterChainResumer is per AsyncFilter. So, yes, the counter will get incremented/decremented per individual filter execution.

PR #419, opened 2018-04-08
Fix: parseRequest lose form-field when value empty

JoneWeng:
We use local-forward in product.

curl -X POST -d 'foo=1&bar='   http://host:port/forward/xxxx

That will lose request parameter  `bar`


artgon:
This PR doesn't actually fix the issue, but I will push a fix for Zuul 2.0 and back port it to Zuul 1.0.

PR #425, opened 2018-04-27
open source origin concurrency protection

artgon:


zuul-core/src/main/java/com/netflix/zuul/origins/BasicNettyOrigin.java
```diff
@@ -71,6 +78,10 @@ public BasicNettyOrigin(String name, String vip, Registry registry) {
         this.clientChannelManager = new DefaultClientChannelManager(name, vip, config, registry);
         this.clientChannelManager.init();
         this.requestAttemptFactory = new NettyRequestAttemptFactory();
+
+        this.concurrentRequests = new AtomicInteger(0);
```
kerumai: Should we add a Gauge metric for this?

Issue #428, opened 2018-05-03
ProxyEndpoint https


JacobJohansen:
Is the proxy endpoint able to handle https endpoints or is a custom endpoint needed?
referencing Zuul version 2.1

artgon:
Currently no, but I'm planning to add it soon.

JacobJohansen:
@artgon were you thinking of using a property along the lines of 

```
this.IS_SECURE = new CachedDynamicBooleanProperty(originName+".netty.client.IsSecure", false);
```

artgon:
@JacobJohansen do you want to submit a PR for this?

JacobJohansen:
I was thinking of creating a pr, but i first wanted to clean up the code formating. review the community guidelines. Lastly I was considering wiring in the jks so that it would have a more expected behavior for custom certs.

PR #435, opened 2018-05-10
GH-428 add client ssl context to suppor secure endpoint proxying

JacobJohansen:


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultOriginChannelInitializer.java
```diff
@@ -82,6 +93,14 @@ protected void initChannel(Channel ch) throws Exception {
     protected void addMethodBindingHandler(ChannelPipeline pipeline) {
     }
 
+    private SslContext getClientSslContext() throws SSLException {
+        return SslContextBuilder
+                    .forClient()
+                    .sslProvider(BaseSslContextFactory.chooseSslProvider())
+                    .startTls(true)
```
artgon: Why enable the `startTls` flag?

JacobJohansen: this triggers the ssl handshake for the connection

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultOriginChannelInitializer.java
```diff
@@ -82,6 +93,14 @@ protected void initChannel(Channel ch) throws Exception {
...
+                    .build();
```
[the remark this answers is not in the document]
artgon: Probably worth configuring the `.ciphers()` explicitly as well. See: https://github.com/Netflix/zuul/blob/2.1/zuul-core/src/main/java/com/netflix/netty/common/ssl/ServerSslConfig.java#L115 and https://github.com/Netflix/zuul/blob/abdc1bd416ccf1fd03c32a12a403ea8de0a84cf0/zuul-core/src/main/java/com/netflix/zuul/netty/ssl/BaseSslContextFactory.java#L164

artgon:
I think maybe I miscommunicated the changes. All the code you had before was fine, just wanted the creation method in the initialized to move to a class `ClientSslContextFactory` that extends `BaseSslContextFactory`. There's no need to create a new interface or copy code to conn pool config. You can reference those static methods I linked to directly, and use the inherited methods directly in your subclass.

artgon:
Let's go back to the original commit, get it merged and we can improve it iteratively.

JacobJohansen:
The rough part in using the baseSslContextFactory is that its tightly coupled to the server configuration. This was my first attempt at separating the server config from an actual base ssl context provider. 

I’ll revert it back to the simpler version when I'm back at my desktop.

JacobJohansen:
@artgon I was able to get in the changes for ciphers and protocols in a simpler way. hope this is the best of both worlds.

artgon:
I think that couple between the `SslConfigs` and `ConnectionPoolImpl` is unnecessary. I agree that those classes can be refactored to be better decoupled but I don't think you need to add it as part of this PR. That's why I wanted the initial, simple implementation you had.

artgon:
This is great. Thanks for your contribution!

PR #436, opened 2018-05-12
Bugfix: multi classes or packages configurations does not work

jingege:
All tests are passed.

### DIFF
--- a/zuul-core/src/test/java/com/netflix/zuul/init2/package-info.java
+++ b/zuul-core/src/test/java/com/netflix/zuul/init2/package-info.java
@@ -1,0 +1,5 @@
+/**
+ * This package is used to test cases that we define multi filter packages in the configuration file.
+ * @author bill 2018-05-13 01:03
+ */
+package com.netflix.zuul.init2;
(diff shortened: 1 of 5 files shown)


artgon:
Not sure what the bug is?

jingege:
@artgon 

If I define two packages like this: zuul.filters.packages=com.pack.a,com.pack.b
zuul2 just scan the first package. The 'com.pack.b' is ignored.

artgon:
Good catch, thanks for the contribution.

zuul-core/src/test/java/com/netflix/zuul/init2/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * This package is used to test cases that we define multi filter packages in the configuration file.
+ * @author bill 2018-05-13 01:03
+ */
+package com.netflix.zuul.init2;
```
artgon: Why is this file necessary?

jingege: No, it's not. It's personal habbit. Should I delete it in another commit?

jingege:
@artgon Done.

PR #438, opened 2018-05-16
Zuul push oss

raksoras:
Open sourcing Zuul Push (WebSocket/SSE) implementation.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushAuthHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushAuthHandler.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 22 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushAuthHandler.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+        else {
+            sendHttpResponse(req, ctx, NOT_FOUND);
+        }
+    }
+
+    protected final Cookies parseCookies(FullHttpRequest req) {
+        final Cookies cookies = new Cookies();
+        final String cookieStr = req.headers().get(HttpHeaderNames.COOKIE);
```
sudheerv: Hmm..doesn't this return a list when there are multiple Cookies involved?

sudheerv:
changes look good to me 👍

zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushAuthHandler.java
raksoras: I believe SET-COOKIE and be multiple but when user agent sends cookies back it is supposed to merge them all in single COOKIE header? This is what is running in production today...

sudheerv: Hmm..yeah, it's a bit unclear..most of the other code in Zuul that deals with Cookies, seems to use _getAll()_ and loop through them.

FWIW, as you said, the RFC does explicitly not allow multiple Cookies in the request..

https://tools.ietf.org/html/rfc6265#section-5.4
.."When the user agent generates an HTTP request, the user agent MUST
   NOT attach more than one Cookie header field...."

HTTP/2 allows multiple Cookie fields via HPACK (for better compression efficiency), but, does state that the HPACK decoder must concatenate all such fields into a single Header before passing it to a HTTP/1.1 context..

https://tools.ietf.org/html/rfc7540#section-8.1.2.5

PR #469, opened 2018-06-21
Retry on READ TIMEOUT only if the HTTP method is idempontent

raksoras:


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -677,7 +677,9 @@ private void handleNoOriginSelected() {
     }
 
     protected boolean isRetryable(final ErrorType err) {
-        if (err == OutboundErrorType.READ_TIMEOUT  || err == OutboundErrorType.RESET_CONNECTION || err == OutboundErrorType.CONNECT_ERROR) {
+        if ((err == OutboundErrorType.RESET_CONNECTION) ||
+            (err == OutboundErrorType.CONNECT_ERROR) ||
+            (err == OutboundErrorType.READ_TIMEOUT && IDEMPOTENT_HTTP_METHODS.contains(zuulRequest.getMethod().toUpperCase()))){
```
artgon: :nitpick: I don't think you need parentheses around each equivalence expression (except the last one).

PR #470, opened 2018-06-21
Respect trailing equals for key-only query params

artgon:


zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -40,17 +40,20 @@
 {
     private final ListMultimap<String, String> delegate;
     private final boolean immutable;
+    private final HashMap<String, Boolean> trailingEquals;
 
     public HttpQueryParams()
     {
         delegate = ArrayListMultimap.create();
         immutable = false;
+        trailingEquals = new HashMap<>();
```
raksoras: Does it need a hash map? There should only ever be at most one param with a trailing "=" Can we just use a member variable, something like "String paramWithTrailingEqaul" and append a "=" at the tail end of the query string if its not null?

artgon: If you take a look at the tests, I have examples of cases where multiple params have mixed and matched trailing `=`.

raksoras: I understood the issue bit differently. The way I interpret, "k=&n=v" really doesn't have a "trailing" =
But I see your point - to faithfully reproduce exact URL you'd need a hash map. carry on.

PR #498, opened 2018-08-18
remove duplicated filter execution summary

jadetang:
When the `ZuulFilterResult` is failed, `FilterProcessor` will add filter execution summary into the context twice.


carl-mastrangelo:
@jadetang Hi, Zuul 1.x is not supported any more.   If you still want to proceed with this PR, Can you add a test to make sure it doesnt break?   (Netflix won't be able to exercise this code path regularly).

PR #535, opened 2018-11-19
add traffic shaping

dyf6372:
add channel traffic shaping for zuul


artgon:
This seems more like a feature that you may want in your own app, rather than something that needs to be part of the core.

dyf6372:
I think traffic shaping is a basic feature of gateway, especially for every channel.

Issue #215, opened 2016-04-07
Netflix Zuul query string encoding


anoll:
This is a copy of the stackoverflow issue:
http://stackoverflow.com/questions/36475845/netflix-zuul-query-string-encoding

When sending a request via Zuul to a client, Zuul seems to change the query String. More specifically, if the client should receive an url-encoded query String, Zuul decodes the query String once. Here is a concrete example:

If "http://localhost:8080/demo/demo?a=http%3A%2F%2Fsomething/" is sent to the client, the client receives as a query String "a=http://something/".

Looking into Zuul`s code, the function "buildZuulRequestQueryParams" uses "HTTPRequestUtils.getInstance().getQueryParams();" which decodes the query String.

Is this a desired feature or a bug?

ausmith:
The implementation of zuul that my group is using had instances where bad query params were sent to it, causing the entire application to error out. https://github.com/Netflix/zuul/pull/161 added the query param decoding, https://github.com/Netflix/zuul/pull/198 handles a bug we found where encoded ampersands were not getting re-encoded, and https://github.com/Netflix/zuul/pull/199 handles cases where query params are missing an `=`.

Hope that answers the "why" for you. As these were merged in, I believe that it is a desired feature.

anoll:
Thanks for the detailed response.

In our case, the client assumes the query string to be encoded in a certain way. Is there a way to configure the behavior of Zuul such that Zuul does not modify the query string? That would be a desired feature for us :-)

ausmith:
@anoll the changes I described were all in the zuul-simple-webapp present in the zuul project, so your custom implementation likely inherited this. If you go into your application's `SimpleHostRoutingFilter.groovy` you should be able to make these modifications yourself.

hohwille:
Zuul simply does too much auto-magic here and therefore some apps stop working if they are operated behind zuul as reverse proxy.
An example is e.g. foswiki that uses semicolon instead of ampersand as separator for query parameters.
See also here for discussions on that but however, the w3c standard seems to allow this:
https://stackoverflow.com/questions/3481664/semicolon-as-url-query-separator
However, zuul does not allow this:
> https://github.com/Netflix/zuul/blob/c87e0c2a6b65a19ca8eb493d281ee812f7b60794/zuul-core/src/main/java/com/netflix/zuul/util/HTTPRequestUtils.java#L166
Ampersand as separator is considered as fixed separator char and semicolon is not honored.
[...]

PR #563, opened 2019-03-25
feature: Control request parameter encoding with configuration

sandy-adi:
This address the problem cited in #215


sandy-adi:
@artgon could you please take a look at this when you get a chance?

sandy-adi:
@artgon build failing with

```
An exception occurred applying plugin request [id: 'nebula.netflixoss', version: '7.1.0']
> Failed to apply plugin [class 'nebula.plugin.bintray.NebulaBintrayPublishingPlugin']
   > Could not create an instance of type nebula.plugin.bintray.BintrayExtension_Decorated.
      > Could not find any public constructor for class nebula.plugin.bintray.BintrayExtension_Decorated which accepts parameters [].
```
not sure why? It appears like an issue with the plugin repo

sandy-adi:
@artgon thoughts on this change?

sandy-adi:
@artgon did you get a chance to look at this PR?

sandy-adi:
@artgon Could you please take a look at this issue and let me know if you have any concerns?

sandy-adi:
Could we please get reviews on this?

mmaio:
This fix would be help for our efforts as well.

rajc28:
We have the same issue , this fix will be really help us.

sandy-adi:
@artgon any thoughts on this PR?

sandy-adi:
@carl-mastrangelo could you please take a look at this?

carl-mastrangelo:
@sandy-adi I'll take a look.   I discussed this briefly with @artgon but want to take a closer look before approving.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -72,12 +78,17 @@ public static HttpQueryParams parse(String queryString) {
...
-                    value = URLDecoder.decode(value, "UTF-8");
-                }
-                catch (Exception e) {
-                    // do nothing
+                if (!propertyFactory.getBooleanProperty(
+                    ZuulConstants.ZUUL_KEEP_ORIGINAL_QUERY_STRING_ENCODING,
+                    false).getValue()
+                ){
+                    try {
+                        name = URLDecoder.decode(name, "UTF-8");
```
carl-mastrangelo: Hmm, I was going to comment on this, but it seems the issue is pre-existing.    The name / value decoding should be done as all or nothing, because otherwise an exception in the value decoder could cause tearing.   

Optional if you want to fix this.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -72,12 +78,17 @@ public static HttpQueryParams parse(String queryString) {
...
                 String value = s.substring(i + 1);
 
-                try {
-                    name = URLDecoder.decode(name, "UTF-8");
-                    value = URLDecoder.decode(value, "UTF-8");
-                }
-                catch (Exception e) {
-                    // do nothing
+                if (!propertyFactory.getBooleanProperty(
+                    ZuulConstants.ZUUL_KEEP_ORIGINAL_QUERY_STRING_ENCODING,
```
carl-mastrangelo: I don't think this is the correct thing to do.   Preserving the encoding is fine, but the name and value still need to be validated.  If this flag is off, all that needs to happen is the return value of URLDecoder should be discarded.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -176,30 +187,34 @@ public void clear()
...
-        StringBuilder sb = new StringBuilder();
-        try {
-            for (Map.Entry<String, String> entry : entries()) {
-                sb.append(URLEncoder.encode(entry.getKey(), "UTF-8"));
-                if (StringUtils.isNotEmpty(entry.getValue())) {
-                    sb.append('=');
-                    sb.append(URLEncoder.encode(entry.getValue(), "UTF-8"));
+        if (propertyFactory.getBooleanProperty(ZuulConstants.ZUUL_KEEP_ORIGINAL_QUERY_STRING_ENCODING,false).getValue()) {
+            return toString();
+        }else {
```
carl-mastrangelo: nit: extra space after `}`

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -176,30 +187,34 @@ public void clear()
...
     public String toEncodedString()
     {
-        StringBuilder sb = new StringBuilder();
-        try {
-            for (Map.Entry<String, String> entry : entries()) {
-                sb.append(URLEncoder.encode(entry.getKey(), "UTF-8"));
-                if (StringUtils.isNotEmpty(entry.getValue())) {
-                    sb.append('=');
-                    sb.append(URLEncoder.encode(entry.getValue(), "UTF-8"));
+        if (propertyFactory.getBooleanProperty(ZuulConstants.ZUUL_KEEP_ORIGINAL_QUERY_STRING_ENCODING,false).getValue()) {
```
carl-mastrangelo: nit: this should not be dynamically changeable.   You should capture the value when the query string is parsed originally, and use that to decide what the string value should be.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -176,30 +187,34 @@ public void clear()
...
                 }
-                sb.append('&');
             }
-
-            // Remove trailing '&'.
-            if (sb.length() > 0 && '&' == sb.charAt(sb.length() - 1)) {
-                sb.deleteCharAt(sb.length() - 1);
+            catch (UnsupportedEncodingException e) {
+                // Won't happen.
+                e.printStackTrace();
```
carl-mastrangelo: Again, I know it's not your fault, but since you are touching this code:   This should be throwing an AssertionError(e), rather than printing to stderr.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -270,6 +285,18 @@ public boolean equals(Object obj)
     @RunWith(MockitoJUnitRunner.class)
     public static class TestUnit
     {
+        Properties testProperties = new Properties();
+
+        private void setKeepOriginalEncoding(boolean keepOriginalEncoding) {
```
carl-mastrangelo: nit: private methods should probably go at the bottom of the class to allow reading the test code from top to bottom.

carl-mastrangelo:
Also, to be honest, I'm not that excited about this change.   Query parameter processing is something that isn't well defined, and different people think it should do different things.   I'm wondering if it would be a better idea to make the HttpQueryParams parsing pluggable, using the existing class as the default?

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
sandy-adi: will add

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
sandy-adi: Based on what has been currently implemented I think it should be a per key/value. Cause in cases where the query string has only key the encoding decoding of the query string varies. I think we should leave it as is.

sandy-adi:
@carl-mastrangelo do you have a better alternative to this? Would you be able to work on it? It will be greatly appreciated if you could fix this problem. Honestly a proxy should not be doing what is happening in this class here.

PR #572, opened 2019-04-23
Ensures jar file is included in publishToMavenLocal

quidryan:
The goal here is to be able to run `./gradlew publishToMavenLocal`, which will enable the possibility of not having `includeBuild` in the internal edgezuul build.

I'll try to follow up with EngTools on why it's needed at all.


kerumai:
@quidryan this seems fine, although we do actually want the `includeBuild` in edgezuul.

PR #590, opened 2019-07-02
move read timeout handling into its own handler

artgon:


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ClientTimeoutHandler.java
```diff
@@ -0,0 +1,76 @@
...
+public class ClientTimeoutHandler extends CombinedChannelDuplexHandler {
+    private static final Logger LOG = LoggerFactory.getLogger(ClientTimeoutHandler.class);
+
+    public static final AttributeKey<Integer> ORIGIN_RESPONSE_READ_TIMEOUT = AttributeKey.newInstance("originResponseReadTimeout");
+
+    public ClientTimeoutHandler() {
+        super(new InboundHandler(), new OutboundHandler());
+    }
+
+    private static class InboundHandler extends ChannelInboundHandlerAdapter {
```
jfernandez: I -think- we can instead extend SimpleChannelInboundHandler<LastHttpContent> to remove the `msg instanceof LastHttpContent` check and not worry about reference counting.

PR #612, opened 2019-09-04
core: Make Server easier to use

carl-mastrangelo:
* Server now attempts to use SocketAddress instead of port number.  This
will make it easie rin the future to adapt it to using other kinds of
addresses, like LocalAddress in tests.
* Automatically tries to select between Epoll, Kqueue and Nio, based on
availablility.  This requires a runtime dependency to be present, and
attempts to avoid loading extra transport specific classses.
* Log levels reduced.  Most of the log statements were not actionable,
and only are useful incase something is going wrong.


carl-mastrangelo:
Tests to come, wanted to get buy in on the approach before continuing.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -57,67 +71,94 @@
...
+     */
+    @Deprecated
+    public static final DynamicBooleanProperty USE_EPOLL =
+            new DynamicBooleanProperty("zuul.server.netty.socket.epoll", false);
+
+    /**
+     * If {@code true}, The Zuul server will avoid autodetecting the transport type and use the default Java NIO
+     * transport.
+     */
+    static final DynamicBooleanProperty forceNio =
```
carl-mastrangelo: I went with lower case here because typically I have seen upper case used for constants. Since this is dynamic, I decided to lower it.    I think generally this pattern should apply, but I don't want to break compat, so I didn't update `USE_EPOLL`.

artgon: Dynamic props are kind of like constants (they rarely change) so that's why you'll see them upper cased in a lot of places.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -57,67 +71,94 @@
...
+        this.serverStatusManager = checkNotNull(serverStatusManager, "serverStatusManager");
+        this.clientConnectionsShutdown = checkNotNull(clientConnectionsShutdown, "clientConnectionsShutdown");
+        this.eventLoopConfig = checkNotNull(eventLoopConfig, "eventLoopConfig");
+        this.eventLoopGroupMetrics = checkNotNull(eventLoopGroupMetrics, "eventLoopGroupMetrics");
     }
 
     public void stop()
     {
-        LOG.warn("Shutting down Zuul.");
+        LOG.debug("Shutting down Zuul.");
```
artgon: Why change these to debug?

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -147,47 +188,40 @@ public void gracefullyShutdownConnections()
...
+        // Apply transport specific socket options.
+        for (Map.Entry<ChannelOption, ?> optionEntry : serverGroup.transportChannelOptions.entrySet()) {
+            serverBootstrap = serverBootstrap.option(optionEntry.getKey(), optionEntry.getValue());
+        }
 
         serverBootstrap.childHandler(channelInitializer);
         serverBootstrap.validate();
 
-        LOG.info("Binding to port: " + port);
+        LOG.debug("Binding to : " + listenAddress);
```
artgon: This is actually a really useful log statement when you have multiple ports being bound.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -147,47 +188,40 @@ public void gracefullyShutdownConnections()
...
-        if (USE_EPOLL.get()) {
-            LOG.warn("Proxy listening with TCP transport using EPOLL");
-            serverBootstrap = serverBootstrap.channel(EpollServerSocketChannel.class);
-            channelOptions.put(EpollChannelOption.TCP_DEFER_ACCEPT, Integer.valueOf(-1));
-        }
-        else {
-            LOG.warn("Proxy listening with TCP transport using NIO");
-            serverBootstrap = serverBootstrap.channel(NioServerSocketChannel.class);
-        }
+        LOG.debug("Proxy listening with " + serverGroup.channelType);
```
artgon: This should be `INFO`.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -242,31 +277,47 @@ private void initializeTransport()
             ThreadFactory workerThreadFactory = new CategorizedThreadFactory(name + "-ClientToZuulWorker");
             Executor workerExecutor = new ThreadPerTaskExecutor(workerThreadFactory);
 
-            if (USE_EPOLL.get()) {
+            Map<ChannelOption, Object> extraOptions = new HashMap<>();
+            boolean useNio = forceNio.get();
```
artgon: Do we want the ability to force other types? i.e. the `USE_EPOLL` flag.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: As mentioned above, they aren't actionable.  Also, it doesn't much matter in the case of JVM shutdown, since the logger is usually gone by that point too.   

I posit that if someone was interested in the shutdown status, we should expose this via a Future, CountDownLatch, or something like that.   The reason is that callers may have their own logging preferences.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: See comment above.   I'd say this isn't actionable, and should be exposed via a programmatic api up to the caller.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: That may be, but again, the caller of this code could log this.    This class is more of a library, and probably should let a higher layer log.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: The DynamicProperties lack the ability to see if they were explicitly set.  `USE_EPOLL` can't be used to see if the caller explicitly didn't want epoll, or if they just didn't care.   The forceNio var is there to re-expose the ability to fall back to a safe transport in case we need it during a production emergency.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
kerumai: Yeah I'd change this to `INFO` (and the other logging of port binding too). It's true to say its not actionable and therefore shouldn't be `WARN`, but I have surprisingly frequently looked at these particular statements in logs to verify that configurations of ports have been applied as intended, so I'd keep them above `DEBUG`.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: Done, though this is still confusing IMO

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: upgraded to info, but I think these should still be logged at a different level

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: Done

carl-mastrangelo:
@artgon PTAL

PR #628, opened 2019-09-27
core:   Don't fire HAProxyMessage further up the pipeline

carl-mastrangelo:
Additionally, don't release the interior TLVs as the channel clode future takes care of that now


carl-mastrangelo:
I also confirmed there are no other handlers in the pipeline that would read this message.

PR #629, opened 2019-09-30
all: move the majority of tests to the test/ directory

carl-mastrangelo:
This is the standard structure for Java projects.    No functionality has been changed as part of this refactor.    I can confirm the number of test cases before and after is 115.


artgon:
Wow, that's a big chunk of work. Thank you! 🙏 🎉 ❤️

PR #630, opened 2019-10-02
zuul-core: lower severity of early SSL handshake failures

carl-mastrangelo:
In newer versions of Netty, failure to write the ServerHello message result in a SslCompletionEvent and the channel closed.   This can happen if the client sends a ClientHello, and then quickly closes the connection.

This change lowers the severity of the message being logged, and adds a test proving the connection cleans up.   This test fails on older versions of netty.

### DIFF
--- a/zuul-core/src/test/java/com/netflix/zuul/netty/server/ssl/SslHandshakeInfoHandlerTest.java
+++ b/zuul-core/src/test/java/com/netflix/zuul/netty/server/ssl/SslHandshakeInfoHandlerTest.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ *      Licensed under the Apache License, Version 2.0 (the "License");
+ *      you may not use this file except in compliance with the License.
+ *      You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *      Unless required by applicable law or agreed to in writing, software
+ *      distributed under the License is distributed on an "AS IS" BASIS,
+ *      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


zuul-core/src/test/java/com/netflix/zuul/netty/server/ssl/SslHandshakeInfoHandlerTest.java
```diff
@@ -0,0 +1,75 @@
...
+public class SslHandshakeInfoHandlerTest {
+
+    @Test
+    public void sslEarlyHandshakeFailure() throws Exception {
+        EmbeddedChannel clientChannel = new EmbeddedChannel();
+        SSLEngine clientEngine = SslContextBuilder.forClient().build().newEngine(clientChannel.alloc());
+        clientChannel.pipeline().addLast(new SslHandler(clientEngine));
+
+        EmbeddedChannel serverChannel = new EmbeddedChannel();
+        SelfSignedCertificate cert = new SelfSignedCertificate("localhorse");
```
artgon: 🐴

zuul-core/src/test/java/com/netflix/zuul/netty/server/ssl/SslHandshakeInfoHandlerTest.java
```diff
@@ -0,0 +1,75 @@
...
+        SSLEngine serverEngine = SslContextBuilder.forServer(cert.key(), cert.cert()).build()
+                .newEngine(serverChannel.alloc());
+
+        serverChannel.pipeline().addLast(new ChannelOutboundHandlerAdapter() {
+            @Override
+            public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+                // Simulate an early closure form the client.
```
artgon: Clever 👍

PR #632, opened 2019-10-04
zuul-core: use non deprecated Http2 Multiplex handlers

carl-mastrangelo:


carl-mastrangelo:
cc: @philrud

zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2OrHttpHandler.java
```diff
@@ -44,7 +43,6 @@
  */
 public class Http2OrHttpHandler extends ApplicationProtocolNegotiationHandler {
     public static final AttributeKey<String> PROTOCOL_NAME = AttributeKey.valueOf("protocol_name");
-    public static final AttributeKey<Http2Connection> H2_CONN_KEY = AttributeKey.newInstance("h2_connection");
```
artgon: What happened to the `H2_CONN_KEY` attribute?

carl-mastrangelo: It was never used internally or externally, so I removed it.   The connection is not accessible in the new API as well.

carl-mastrangelo:
The origin side will follow up shortly.

PR #638, opened 2019-10-17
Modify a simple bug for KEY_EVENT_PROPS

linwenyao:
The value of key: ZUUL_ORIGIN_CHOSEN_HOST_ADDR_MAP_KEY should be attempToChosenHostMap instead of attempToIpAddressMap.


carl-mastrangelo:
@linwenyao Merged, thanks!

PR #640, opened 2019-10-18
zuul-core: add perfmark annotations

carl-mastrangelo:
I tried to keep this as clean as possible.  PerfMark does make things more verbose, but it should have almost no effect on the runtime performance.  The bridge between the async observable and observe is slightly annoying, but I believe works properly.   In particular, making the links available to the observer from the observable appears to have a happens-before relationship in the Rx implementation.  I used compareAndSet to make sure it is only set once, incase onNext is accidentally called twice.

Tested manually, it looks like it is working properly.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/ZuulFilter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/ZuulFilter.java
@@ -18,6 +18,8 @@ package com.netflix.zuul.filters;
 import com.netflix.zuul.exception.ZuulFilterConcurrencyExceededException;
 import com.netflix.zuul.message.ZuulMessage;
 import io.netty.handler.codec.http.HttpContent;
+import io.perfmark.PerfMark;
+import io.perfmark.Tag;
 import rx.Observable;
 
 /**
@@ -98,4 +100,14 @@ public interface ZuulFilter<I extends ZuulMessage, O extends ZuulMessage> extend
      * @return
      */
     HttpContent processContentChunk(ZuulMessage zuulMessage, HttpContent chunk);
+
+    /**
+     * Returns the internal profiling tag for this filter.   This method is NOT API stable, and may be removed.
+     * This method is internal.
+     * @return
+     */
+    default Tag perfmarkTag() {
+        return PerfMark.createTag();
+    }
+
 }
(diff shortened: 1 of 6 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -171,18 +180,66 @@ protected final O filter(final ZuulFilter<I, O> filter, final I inMesg) {
...
+                        })
+                        .doOnCompleted(() -> {
+                            PerfMark.startTask("FilterRunner.onCompletedAsync", filter.perfmarkTag());
+                            try {
+                                PerfMark.linkIn(nettyToSchedulerLink);
+                                schedulerToNettyLinks.compareAndSet(2, null, PerfMark.linkOut());
+                            } finally {
+                                PerfMark.stopTask("FilterRunner.onCompletedAsync", filter.perfmarkTag());
+                            }
+                        })
```
artgon: Would be nice to move all of these out of this method and into a wrapper. It creates confusion around what the code is doing vs what the benchmarks are doing.

zuul-core/src/main/java/com/netflix/zuul/filters/ZuulFilter.java
```diff
@@ -98,4 +100,14 @@
      * @return
      */
     HttpContent processContentChunk(ZuulMessage zuulMessage, HttpContent chunk);
+
+    /**
+     * Returns the internal profiling tag for this filter.   This method is NOT API stable, and may be removed.
+     * This method is internal.
+     * @return
+     */
+    default Tag perfmarkTag() {
```
artgon: Doesn't adding this method here, remove the need to add it to all the other interfaces?

carl-mastrangelo: This method is for the callers (in the core fo Zuul), but it just returns a dummy tag that has no information.    This is just to make it safe to call on any subclasses, incase they dont yet implement this method (i.e. code outside of ours).    The sub classes what to implement this method to return their named tags, which are cached on each filter instance.

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
carl-mastrangelo: Point taken, refactored

artgon:
LGTM

PR #642, opened 2019-10-18
Prepare BasicNettyOrigin for subclassing

steffentemplin:
BasicNettyOrigin contains some valuable logic like status recording and concurrency protection. Therefore it would be an ideal base class for more complex NettyOrigin implementations. This PR introduces the following changes:

1. Make relevant fields protected
2. Separate construction from initializing the downstream objects like `IClientConfig` and `ClientChannelManager` by introducing a post-construct method`init()`
3. Introduce method to selectively override `ClientChannelManager` construction
4. Protected method to obtain an LB server selection key based on the Zuul request object


artgon:
This class is intended to be more of a reference implementation than something that is subclassed. I think a better approach would be to create your own subclass of `NettyOrigin` that's specific to your needs.

steffentemplin:
Fair enough, I had given it a 50/50 chance of being accepted :) Closed.

PR #648, opened 2019-10-23
zuul-core:  Make BaseServerStartup prefer using socket addresses

carl-mastrangelo:
This is an incremental change to allow existing code to slowly move over to socket addresses rather than ports.   Old clients can still override existing methods (but with a deprecation warning), while new clients can override the socket address methods.

Updates CGW-95

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
@@ -43,6 +43,7 @@ import io.netty.util.concurrent.DefaultEventExecutorChooserFactory;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.EventExecutorChooserFactory;
 import io.netty.util.concurrent.ThreadPerTaskExecutor;
+import java.util.LinkedHashMap;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -116,11 +127,11 @@ public Server(Map<Integer, ChannelInitializer> portsToChannelInitializers, Serve
                 clientConnectionsShutdown, eventLoopGroupMetrics, eventLoopConfig);
     }
 
-    Server(ServerStatusManager serverStatusManager,
+    protected Server(ServerStatusManager serverStatusManager,
            Map<? extends SocketAddress, ? extends ChannelInitializer<?>> addressesToInitializers,
            ClientConnectionsShutdown clientConnectionsShutdown, EventLoopGroupMetrics eventLoopGroupMetrics,
            EventLoopConfig eventLoopConfig) {
-        this.addressesToInitializers = Collections.unmodifiableMap(new HashMap<>(addressesToInitializers));
+        this.addressesToInitializers = Collections.unmodifiableMap(new LinkedHashMap<>(addressesToInitializers));
```
carl-mastrangelo: while not strictly necessary, ordering the hashmap makes future testing easier

PR #650, opened 2019-10-24
zuul-core: make BaseZuulChannelInitializer prefer a string over port

carl-mastrangelo:
Observation: the port number is only ever used to do metric recording, and not for anything networking related.   This change overloads the `BaseZuulChannelInitializer` constructor to prefer a "metric suffix" instead of the port, and deprecates the field.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/BaseZuulChannelInitializer.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/BaseZuulChannelInitializer.java
@@ -70,7 +70,6 @@ import io.netty.util.AttributeKey;
 
 import java.util.List;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/BaseZuulChannelInitializer.java
```diff
@@ -71,7 +70,9 @@
 
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
```
artgon: unused?

carl-mastrangelo: Holdover from the uglier predecessor of this PR.  Removed!

PR #652, opened 2019-10-24
all: update most of the initializers to use metricSuffix

carl-mastrangelo:


artgon:
As discussed in person, minor nit is that you're setting the "id" tag of the metric -- so should indicate that in the parameter somehow (e.g. metricId or metricIdSuffix).

zuul-core/src/main/java/com/netflix/zuul/netty/server/ZuulServerChannelInitializer.java
```diff
@@ -28,12 +28,24 @@
...
+    /**
+     * Use {@link #ZuulServerChannelInitializer(String, ChannelConfig, ChannelConfig, ChannelGroup)} instead.
+     */
+    @Deprecated
+    public ZuulServerChannelInitializer(
+            int port,
+            ChannelConfig channelConfig,
+            ChannelConfig channelDependencies,
+            ChannelGroup channels) {
+        this(String.valueOf(port), channelConfig, channelDependencies, channels);
```
artgon: Maybe add a convenience constructor to `BaseZuulChannelInitializer` that takes an `int`?

carl-mastrangelo:
Id used instead of suffix

zuul-core/src/main/java/com/netflix/zuul/netty/server/ZuulServerChannelInitializer.java
carl-mastrangelo: I'm trying to stamp out all uses of the int param, so passing the string makes it easier for me to see that it's correct.   I would prefer updating this later, since it lets me ensure all the correct places have been updated.

PR #659, opened 2019-10-29
zuul-core: support specific bind addresses

carl-mastrangelo:


carl-mastrangelo:
I don't think we will use UDS any time soon, but I just wanted to show that this is flexible enough to support it.

I have not yet included support for "only" one kind of address type, since this involves setting some custom socket options (namely `IPV6_V6ONLY `), which will probably need some special wiring.  

Lastly, for ease of specification, I allow leaving the key blank, since it's a lot easier to put in configuration `=7001`, rather than include the `ANY`.

carl-mastrangelo:
Also, the weird switch statements are there to help with static analysis.   I plan on eventually adding ErrorProne checks to Zuul.  This makes it hard to forget to update the code when new enum values are added, since it will be a compile time failure.

zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
```diff
@@ -0,0 +1,173 @@
...
+ *         <li>{@code ANY_LOCAL=7001} Binds on localhost for all IP stacks for port 7001</li>
+ *         <li>{@code IPV4_LOCAL=7001} Binds on IPv4 localhost (127.0.0.1) for port 7001</li>
+ *         <li>{@code IPV6_LOCAL=7001} Binds on IPv6 localhost (::1) for port 7001</li>
+ *         <li>{@code UDS=/var/run/zuul.sock} Binds a domain socket at /var/run/zuul.sock</li>
+ *     </ul>
+ *
+ *  <p>Note that the local IPv4 binds only work for {@code 127.0.0.1}, and not any other loopback addresses.  Currently,
+ *      all IP stack specific bind types only "prefer" a stack; it is up to the OS and the JVM to pick the the final
+ *      address.
+ */
```
artgon: This is great, thank you for the detailed explanation.

zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
```diff
@@ -0,0 +1,173 @@
...
+        /**
+         * Binds on the IPv6 {@code ::1} localhost address.
+         */
+        IPV6_LOCAL(() ->
+                InetAddress.getByAddress("localhost", new byte[] {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1})),
+        /**
+         * Binds on the Unix Domain Socket path.
+         */
+        UDS,
+        ;
```
artgon: Great comments 🙇 👍

zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
```diff
@@ -0,0 +1,173 @@
...
+            if (input == null || input.isEmpty()) {
+                throw new IllegalArgumentException("Invalid address");
+            }
+
+            int equalsPosition = input.indexOf('=');
+            if (equalsPosition == -1) {
+                throw new IllegalArgumentException("Invalid address " + input);
+            }
+            String rawBindType = equalsPosition != 0 ? input.substring(0, equalsPosition) : BindType.ANY.name();
+            BindType bindType = BindType.valueOf(rawBindType.toUpperCase(Locale.ROOT));
```
artgon: Why add Locale here?

zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
```diff
@@ -0,0 +1,173 @@
...
+        public SocketAddress apply(String input) {
+            if (input == null || input.isEmpty()) {
+                throw new IllegalArgumentException("Invalid address");
+            }
+
+            int equalsPosition = input.indexOf('=');
+            if (equalsPosition == -1) {
+                throw new IllegalArgumentException("Invalid address " + input);
+            }
+            String rawBindType = equalsPosition != 0 ? input.substring(0, equalsPosition) : BindType.ANY.name();
```
artgon: Why not `split` on `=`?

zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
carl-mastrangelo: If not specific, the behavior changes based on the system you run on.  Picking ROOT forces the behavior to be consistent.  (for example, if for some reason in the future unicode was supported here).    Also, ErrorProne will nag if you don't.

zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
carl-mastrangelo: I wanted to the behavior to only split on the first `=`, rather than all of them, that way there can be `=` in the value portion.

Additionally, I have been scared off from using split:   See the docs for Guava's Splitter: https://github.com/google/guava/wiki/StringsExplained#splitter

PR #664, opened 2019-11-01
zuul-core: remove ability to set port on HttpRequestMessage

carl-mastrangelo:


carl-mastrangelo:
I am going to be moving most of the port/IP methods on ZuulMessage subclasses to use a socket address, and special case InetSocketAddress where the port/IP values are used.

This PR is moving the code away from being able to set the values.  There will be multiple followup PRs.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessage.java
```diff
@@ -31,12 +31,8 @@
 
     void setPath(String path);
 
-    void setClientIp(String clientIp);
```
carl-mastrangelo: This is an API breaking change

PR #684, opened 2019-11-26
all: remove more references to the word 'port'

carl-mastrangelo:


carl-mastrangelo:
The main benefit is that it lets me keep track of which ports I have converted over to SocketAddress.

PR #686, opened 2019-12-04
Add docs about source addresses

carl-mastrangelo:


carl-mastrangelo:
In the near future, I am going to steer Zuul away from InetSocketAddress and port numbers in the downstream handlers.   This change is about documenting existing code and behaviors, so that I can refactor safely later.

PR #687, opened 2019-12-04
all: Improve README - specific description for Zuul

argha-c:


README.md
```diff
@@ -4,7 +4,7 @@
 
 [![Build Status](https://travis-ci.org/Netflix/zuul.svg)](https://travis-ci.org/Netflix/zuul/builds)
 
-Zuul is an edge service that provides dynamic routing, monitoring, resiliency, security, and more.
+Zuul is an L7 gateway that provides capabilties for dynamic routing, monitoring, resiliency, security, and more.
```
carl-mastrangelo: s/capabilties/capabilities/

artgon: I think "application gateway" would be clearer than "L7 gateway".

argha-c: I would like to be explicit about L7. "Application" should capture that in an ideal world but does "L7 application gateway" work?

argha-c: The insistence is because application/API gateways sometimes get used interchangeably, and the scope is broader than what Zuul is meant to do. [Here's a good discussion](https://blog.christianposta.com/microservices/api-gateways-are-going-through-an-identity-crisis/)

argha-c: Fixed.

artgon: Can you clarify the distinction, wasn't clear in the blog post? Why is L7 more important to state than the other words?

argha-c: Application/API gateways do more than Zuul is meant for IMHO, that may include aggregation, interface bridge etc.
L7 proxies are closer in terms of capability mapping for what Zuul does but nothing limits Zuul to being used only as service mesh?

PR #688, opened 2019-12-04
zuul-core: rework Cookie logic to use non-deprecated types

carl-mastrangelo:
This does a few things:

* Deprecates most of the cookie methods on the `Cookie` class
* Adds test for the Cookie class
* Updates the PushAuthHandler to use non-deprecated types
* adds tests for push.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
@@ -16,55 +16,108 @@
 
 package com.netflix.zuul.message.http;
 
+import static java.util.Collections.unmodifiableList;
+import static java.util.stream.Collectors.toList;
+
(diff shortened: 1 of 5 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
```diff
@@ -16,62 +16,127 @@
...
- * Date: 6/18/15
- * Time: 12:04 AM
+ * A container for cookies.
  */
-public class Cookies
-{
-    private Map<String, List<Cookie>> map = new HashMap<>();
-    private List<Cookie> all = new ArrayList<>();
+public final class Cookies {
+    private final Map<String, List<io.netty.handler.codec.http.cookie.Cookie>> map = new LinkedHashMap<>();
```
artgon: Why use the full package name?

PR #689, opened 2019-12-06
zuul-core,zuul-sample: avoid using deprecated cookie apis

carl-mastrangelo:


artgon:
Changing the default to strict may break users. Why not keep the same default as the original encoder/decoder?

carl-mastrangelo:
@artgon I tried to keep it pretty close to the original.   The false in a couple places I believe corresponds to lax, and the lack of the second param means strict.    Could you point to the specific line?

carl-mastrangelo:
Based on offline conversation, I think this preserves decoding existing behavior.   Only encoding has become strict, but since we don't encode headers from origins or clients (only the ones we add in), this should be okay.

@artgon SG?

PR #698, opened 2020-01-03
zuul-core: support listing listening addresses

carl-mastrangelo:


zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -272,6 +280,12 @@ public void postEventLoopCreationHook(EventLoopGroup clientToProxyBossPool, Even
 
         private volatile boolean stopped = false;
 
+        private final List<Channel> nettyServers = new ArrayList<>();
```
argha-c: Can this be a `Set` instead?

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -287,6 +301,17 @@ public void uncaughtException(final Thread t, final Throwable e) {
             Runtime.getRuntime().addShutdownHook(new Thread(() -> stop(), "Zuul-ServerGroup-JVM-shutdown-hook"));
         }
 
+        synchronized List<SocketAddress> getListeningAddresses() {
```
argha-c: s/List/Set ?

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: Sure, though I don't think it matters

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: I don't think this can be, since equality on SocketAddresses is not well defined.   (resolved InetSocketAddresses vs unresolved ones)

zuul-core/src/test/java/com/netflix/zuul/netty/server/ServerTest.java
```diff
@@ -0,0 +1,82 @@
...
+ * Tests for {@link Server}.
+ */
+@RunWith(JUnit4.class)
+public class ServerTest {
+
+    @Test
+    public void getListeningSockets() {
+        ServerStatusManager ssm = mock(ServerStatusManager.class);
+        Map<SocketAddress, ChannelInitializer<?>> initializers = new HashMap<>();
+        initializers.put(
```
argha-c: Might want to init with at least 2 addrs. Weird things manifest themselves in collections of a single size 😄

carl-mastrangelo: Good point, done.

PR #704, opened 2020-01-16
zuul-core: Include additional client address context in `HttpRequestMessage`

argha-c:
We set the channel attribute `ATTR_LOCAL_ADDR` [based on the LB proxy protocol config here](https://github.com/Netflix/zuul/blob/master/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/ElbProxyProtocolChannelHandler.java#L132), which should also work for IPv6 addresses.
This change should be especially relevant when proxy protocol is enabled.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestInfo.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestInfo.java
@@ -18,6 +18,7 @@ package com.netflix.zuul.message.http;
 
 import com.netflix.zuul.message.Headers;
 import com.netflix.zuul.message.ZuulMessage;
+import java.util.Optional;
 
 /**
  * User: Mike Smith
@@ -58,8 +59,13 @@ public interface HttpRequestInfo extends ZuulMessage
 
     int getOriginalPort();
 
-    default int getClientDestinationPort() {
-        return getOriginalPort();
+    /**
+     * Reflects the actual destination port that the client intended to communicate with,
+     * in preference to the port Zuul was listening on. In the case where proxy protocol is
+     * enabled, this should reflect the destination IP encoded in the TCP payload by the load balancer.
+     */
+    default Optional<Integer> getClientDestinationPort() {
+        throw new UnsupportedOperationException();
     }
 
     String reconstructURI();
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestInfo.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestInfo.java
@@ -58,6 +58,10 @@ public interface HttpRequestInfo extends ZuulMessage
 
     int getOriginalPort();
 
+    default int getClientDestinationPort() {
+        return getOriginalPort();
+    }
+
     String reconstructURI();
 
     /** Parse and lazily cache the request cookies. */
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
@@ -30,6 +30,8 @@ import com.netflix.zuul.util.HttpUtils;
 import io.netty.handler.codec.http.Cookie;
 import io.netty.handler.codec.http.CookieDecoder;
 import io.netty.handler.codec.http.HttpContent;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -83,6 +85,7 @@ public class HttpRequestMessageImpl implements HttpRequestMessage
     private String scheme;
     private int port;
     private String serverName;
+    private SocketAddress clientRemoteAddress;
 
     private HttpRequestInfo inboundRequest = null;
     private Cookies parsedCookies = null;
@@ -95,14 +98,14 @@ public class HttpRequestMessageImpl implements HttpRequestMessage
 
     public HttpRequestMessageImpl(SessionContext context, String protocol, String method, String path,
                                   HttpQueryParams queryParams, Headers headers, String clientIp, String scheme,
-                                  int port, String serverName)
+                                  int port, String serverName, SocketAddress clientRemoteAddress)
     {
-        this(context, protocol, method, path, queryParams, headers, clientIp, scheme, port, serverName, false);
+        this(context, protocol, method, path, queryParams, headers, clientIp, scheme, port, serverName, clientRemoteAddress, false);
     }
 
     public HttpRequestMessageImpl(SessionContext context, String protocol, String method, String path,
                                   HttpQueryParams queryParams, Headers headers, String clientIp, String scheme,
-                                  int port, String serverName,
+                                  int port, String serverName, SocketAddress clientRemoteAddress,
                                   boolean immutable)
     {
         this.immutable = immutable;
@@ -123,6 +126,7 @@ public class HttpRequestMessageImpl implements HttpRequestMessage
         this.scheme = scheme;
         this.port = port;
         this.serverName = serverName;
+        this.clientRemoteAddress = clientRemoteAddress;
     }
 
     private void immutableCheck()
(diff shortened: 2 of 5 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestInfo.java
```diff
@@ -58,6 +58,10 @@
 
     int getOriginalPort();
 
+    default int getClientDestinationPort() {
```
carl-mastrangelo: This should throw an unsupported operation exception.  This method also needs a Javadoc.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -123,6 +126,7 @@ public HttpRequestMessageImpl(SessionContext context, String protocol, String me
         this.scheme = scheme;
         this.port = port;
         this.serverName = serverName;
+        this.clientRemoteAddress = clientRemoteAddress;
```
carl-mastrangelo: checkNotNull()

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -528,6 +532,15 @@ public int getOriginalPort()
         return port;
     }
 
+    @Override
+    public int getClientDestinationPort() {
```
carl-mastrangelo: This may be absent, so the API should express this.   It can either be an `Optional<Integer>` or a `@Nullable Integer`

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -254,6 +255,7 @@ private HttpRequestMessage buildZuulHttpRequest(final HttpRequest nativeRequest,
         // This is the only way I found to get the port of the request with netty...
         final int port = channel.attr(SourceAddressChannelHandler.ATTR_SERVER_LOCAL_PORT).get();
         final String serverName = channel.attr(SourceAddressChannelHandler.ATTR_SERVER_LOCAL_ADDRESS).get();
+        final SocketAddress clientDestinationAddress = channel.attr(SourceAddressChannelHandler.ATTR_LOCAL_ADDR).get();
```
carl-mastrangelo: Is this attribute guaranteed to be set?

zuul-core/src/test/java/com/netflix/zuul/context/DebugTest.java
```diff
@@ -59,7 +60,7 @@ public void setup() {
         params.add("k1", "v1");
 
         request = new HttpRequestMessageImpl(ctx, "HTTP/1.1", "post", "/some/where",
-            params, headers, "9.9.9.9", "https", 80, "localhost");
+            params, headers, "9.9.9.9", "https", 80, "localhost",new InetSocketAddress("api.netflix.com", 443));
```
carl-mastrangelo: This is not ideal, because it will do a DNS lookup as part of the test, making it possibly more flaky, and slower.

zuul-core/src/test/java/com/netflix/zuul/message/http/HttpRequestMessageImplTest.java
```diff
@@ -43,7 +45,7 @@ public void testOriginalRequestInfo() {
         Headers headers = new Headers();
         headers.add("Host", "blah.netflix.com");
         request = new HttpRequestMessageImpl(new SessionContext(), "HTTP/1.1", "POST", "/some/where", queryParams, headers,
-            "192.168.0.2", "https", 7002, "localhost");
+            "192.168.0.2", "https", 7002, "localhost", new InetSocketAddress("api.netflix.com", 443));
```
carl-mastrangelo: Same comment

zuul-core/src/test/java/com/netflix/zuul/message/http/HttpRequestMessageImplTest.java
```diff
@@ -239,4 +241,13 @@ public void testCleanCookieHeaders() {
         assertEquals(" BlahId=12345; something=67890;", HttpRequestMessageImpl.cleanCookieHeader(" Secure, BlahId=12345; Secure, something=67890;"));
         assertEquals("", HttpRequestMessageImpl.cleanCookieHeader(""));
     }
+
+    @Test
+    public void shouldPreferClientDestPortWhenInitialized() {
```
carl-mastrangelo: Line wrapping is odd here.    Either fill all the lines all the way, or put one arg per line.  This is for consistency.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
argha-c: I chose to fallback to the implementation for `getOriginalPort`. In that case, the port should not have those values.
We could modify the API to `Optional` if that makes more sense, as opposed to a best effort port value.

zuul-core/src/test/java/com/netflix/zuul/message/http/HttpRequestMessageImplTest.java
argha-c: Yes, I'll actually reformat with our code style formatter here.

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
argha-c: Not really, but the getter's API now indicates an optional value, which should be sufficient.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
argha-c: Not sure we'd want to enforce this knowing the attribute might be absent.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestInfo.java
argha-c: Done.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -528,6 +533,16 @@ public int getOriginalPort()
         return port;
     }
 
+    @Override
+    public Optional<Integer> getClientDestinationPort() {
```
artgon: Don't think `Optional` is supposed to be used as a generic return type.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestInfo.java
```diff
@@ -58,6 +59,15 @@
 
     int getOriginalPort();
 
+    /**
+     * Reflects the actual destination port that the client intended to communicate with,
+     * in preference to the port Zuul was listening on. In the case where proxy protocol is
+     * enabled, this should reflect the destination IP encoded in the TCP payload by the load balancer.
+     */
+    default Optional<Integer> getClientDestinationPort() {
```
artgon: Same as other comment. Should use actual type rather than generic `Optional`.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
argha-c: Not sure I understand this. `Optional` is not recommended as a parameter type
or in substitution for empty collections. I'm not aware of reasons for not using it for response types.

artgon: Yeah, just an opinion, feel free to ignore.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
carl-mastrangelo: If it's really possible to be absent, then it needs to be marked `@Nullable`

zuul-core/src/test/java/com/netflix/zuul/context/DebugTest.java
```diff
@@ -59,7 +61,7 @@ public void setup() {
         params.add("k1", "v1");
 
         request = new HttpRequestMessageImpl(ctx, "HTTP/1.1", "post", "/some/where",
-            params, headers, "9.9.9.9", "https", 80, "localhost");
+            params, headers, "9.9.9.9", "https", 80, "localhost",new LocalAddress("777"));
```
carl-mastrangelo: nit: missing space

carl-mastrangelo: optional: You can also use an anonymous class: `new SocketAddress() {}`

PR #711, opened 2020-01-22
Use server cookie decoder

sunnysingh85:
* Restructure the code to move the parsing logic from headers into the Cookies class. This will make us not refer to the request class for parsing cookies from the headers.
* Use ServerCookieDecoder which now handles multiple cookies with the same name.
* Add tests to make sure that the methods in Cookies class work as expected.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
@@ -16,12 +16,19 @@
 
 package com.netflix.zuul.message.http;
 
+import com.netflix.config.CachedDynamicBooleanProperty;
+import com.netflix.config.DynamicStringProperty;
+import com.netflix.zuul.message.Headers;
 import io.netty.handler.codec.http.Cookie;
+import io.netty.handler.codec.http.DefaultCookie;
+import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * User: Mike Smith
@@ -30,6 +37,23 @@ import java.util.Map;
  */
 public class Cookies
 {
+    private static final Logger LOG = LoggerFactory.getLogger(Cookies.class);
+
+    private static final CachedDynamicBooleanProperty CLEAN_COOKIES = new CachedDynamicBooleanProperty(
+            "zuul.HttpRequestMessage.cookies.clean", false);
+
+    private static final List<Pattern> RE_STRIP;
+    /** ":::"-delimited list of regexes to strip out of the cookie headers. */
+    private static final DynamicStringProperty REGEX_PTNS_TO_STRIP_PROP =
+            new DynamicStringProperty("zuul.request.cookie.cleaner.strip", " Secure,");
+
+    static {
+        RE_STRIP = new ArrayList<>();
+        for (String ptn : REGEX_PTNS_TO_STRIP_PROP.get().split(":::")) {
+            RE_STRIP.add(Pattern.compile(ptn));
+        }
+    }
+
     private Map<String, List<Cookie>> map = new HashMap<>();
     private List<Cookie> all = new ArrayList<>();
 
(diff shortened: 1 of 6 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
```diff
@@ -16,12 +16,19 @@
...
+import io.netty.handler.codec.http.DefaultCookie;
+import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
carl-mastrangelo: The style guide we use suggests not using wildcar imports.

zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
```diff
@@ -30,6 +37,23 @@
...
 public class Cookies
 {
+    private static final Logger LOG = LoggerFactory.getLogger(Cookies.class);
+
+    private static final CachedDynamicBooleanProperty CLEAN_COOKIES = new CachedDynamicBooleanProperty(
+            "zuul.HttpRequestMessage.cookies.clean", false);
+
+    private static final List<Pattern> RE_STRIP;
+    /** ":::"-delimited list of regexes to strip out of the cookie headers. */
+    private static final DynamicStringProperty REGEX_PTNS_TO_STRIP_PROP =
```
carl-mastrangelo: Long term I think this needs to be removed, because it is a hack in the first place.    I think it also doesn't work with the other values like `Expires` which sometimes get through.

Can you mark this `@Deprecated` with a comment on the intent to remove?

zuul-core/src/test/java/com/netflix/zuul/message/http/CookiesTest.java
```diff
@@ -0,0 +1,145 @@
...
+import com.netflix.zuul.message.Headers;
+import io.netty.handler.codec.http.Cookie;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.util.List;
+
+import static com.netflix.zuul.message.http.Cookies.cleanCookieHeader;
+import static org.junit.Assert.*;
```
carl-mastrangelo: please use flat imports.    We have a codestyle xml file you can use in the internal repo.

zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
sunnysingh85: Addressed.

zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
sunnysingh85: Done.

zuul-core/src/test/java/com/netflix/zuul/message/http/CookiesTest.java
sunnysingh85: Addressed.

zuul-core/src/test/java/com/netflix/zuul/message/http/CookiesTest.java
```diff
@@ -0,0 +1,132 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import com.netflix.zuul.message.Headers;
+import io.netty.handler.codec.http.Cookie;
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
```
argha-c: This should use the `JUnit4` runner.

zuul-core/src/main/java/com/netflix/zuul/message/http/CookiesSanitizer.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Performs a simple cleanup of the cookie headers by removing values which shouldn't be part of a cookie header.
+ * Note: It's a temporary hack put in place to clean cookie headers which will removed in future releases.
+ */
+@Deprecated
+class CookiesSanitizer {
+    private static final CachedDynamicBooleanProperty CLEAN_COOKIES = new CachedDynamicBooleanProperty(
+            "zuul.HttpRequestMessage.cookies.clean", false);
```
argha-c: nit: I think the naming convention for FPs is all lower case.

zuul-core/src/test/java/com/netflix/zuul/message/http/CookieSanitizerTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static com.netflix.zuul.message.http.CookiesSanitizer.cleanCookieHeader;
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CookieSanitizerTest {
+
+    @Test
```
argha-c: Can we add a test here that also verifies that the class parses the additional patterns injected via FP correctly, ie. logic in the static initialization block?

zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
```diff
@@ -16,25 +16,56 @@
...
-    public void add(Cookie cookie)
-    {
+    public static Cookies fromHeaders(Headers headers) {
+        Cookies cookies = new Cookies();
+        for (String aCookieHeader : headers.get(HttpHeaderNames.COOKIE)) {
+            try {
+                if (CLEAN_COOKIES.get()) {
+                    aCookieHeader = CookiesSanitizer.cleanCookieHeader(aCookieHeader);
+                }
+                List<io.netty.handler.codec.http.cookie.Cookie> decodedCookies =
```
argha-c: While I understand this isn't directly part of your change, it'd be nice if we could take this opportunity to encapsulate this logic to `Cookies` instead.
Keeping the deprecated `Cookie` local to that class, consider adding a utility method that can do the mapping given a collection of `io.netty.handler.codec.http.cookie.Cookie`?

PR #717, opened 2020-01-28
ZuulFiltersModule now injects AbstractConfiguration.

### LINKED ISSUE #716: Zuul Filters Configuration Injection
Hey Zuul Crew,

The [ZuulFiltersModule](https://github.com/Netflix/zuul/blob/master/zuul-core/src/main/java/com/netflix/zuul/init/ZuulFiltersModule.java#L65) statically loads an `AbstractConfiguration` instance from `ConfigurationManager.getConfigInstance()`. This causes some issues trying to integrate this back into the Netflix environment as Archaius blows up intentionally when `ConfigurationManager` is statically called.

Would you be opposed to injecting an `AbstractConfiguration` instead, which would allow the OSS version to continue to work with `ConfigurationManager` and anything that loads it internally to work with `NetflixConfiguration`?

codyrioux:
ZuulFiltersModule used to statically load AbstractConfiguration via
ConfigurationManager.getConfigInstance(). This limits the user's ability
to bind various configuration loading mechanisms via the injector.

This fixes #716

### DIFF
--- a/zuul-core/src/test/java/com/netflix/zuul/init/InitTestModule.java
+++ b/zuul-core/src/test/java/com/netflix/zuul/init/InitTestModule.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ *      Licensed under the Apache License, Version 2.0 (the "License");
+ *      you may not use this file except in compliance with the License.
+ *      You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *      Unless required by applicable law or agreed to in writing, software
+ *      distributed under the License is distributed on an "AS IS" BASIS,
+ *      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *      See the License for the specific language governing permissions and
+ *      limitations under the License.
+ */
+
+package com.netflix.zuul.init;
+
+import com.google.inject.AbstractModule;
+import com.netflix.config.ConfigurationManager;
+import org.apache.commons.configuration.AbstractConfiguration;
+
+public class InitTestModule extends AbstractModule {
+    @Override
+    protected void configure() {
+        bind(AbstractConfiguration.class).toInstance(ConfigurationManager.getConfigInstance());
+    }
+
+}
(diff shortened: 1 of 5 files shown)


zuul-sample/src/main/java/com/netflix/zuul/sample/ZuulSampleModule.java
```diff
@@ -45,6 +47,13 @@
 public class ZuulSampleModule extends AbstractModule {
     @Override
     protected void configure() {
+        try {
+          ConfigurationManager.loadCascadedPropertiesFromResources("application");
+        } catch (Exception ex) {
+
```
carl-mastrangelo: this should re throw as a RuntimeException

codyrioux: Ooops I mentally noted I'd need to do something with that catch block when I left it empty. Thanks for catching that. Updated.

zuul-sample/src/main/java/com/netflix/zuul/sample/ZuulSampleModule.java
```diff
@@ -45,6 +47,13 @@
 public class ZuulSampleModule extends AbstractModule {
     @Override
     protected void configure() {
+        try {
+          ConfigurationManager.loadCascadedPropertiesFromResources("application");
+        } catch (Exception ex) {
+          throw new RuntimeException("Error loading configuration: " + ex.getMessage());
```
carl-mastrangelo: Not to be nitpicky, but what I meant was that it needs to chain the exceptions together, to make sure the original stack trace isn't lost.

zuul-core/src/test/java/com/netflix/zuul/init/InitTestModule.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2018 Netflix, Inc.
```
carl-mastrangelo: 2020

codyrioux:
Got those last two small changes implemented, and rewrote history so that a merge won't muddy the master commit log.

carl-mastrangelo:
Merged, @codyrioux  thanks!

PR #718, opened 2020-01-31
Refactor eureka related classes to separate module

asafalima:
We wanted to remove the need for eureka from zuul-core for few reasons:
- We don't use it in our organisation, and like us there are others which use different service discovery solution.
- Eureka uses Jersey 1.x and JAX-RS 1.x which conflicts with our rest-sever which is based on Jersey 2.x
- Avoid unnecessary dependencies

To accomplish this goal I made the following changes:
- Created a new module `zuul-eureka` and moved any eureka related file to it.
- Added `ServerStatusManager` interface in core module and `EurekaServerStatusManager` implementation in the eureka module.
- Added `ServerIpAddrExtractor` interface in core module and `DiscoveryServerIpAddrExtractor` implementation in the eureka module which extracts the IP address from `DiscoveryEnabledServer`.
- Refactored out the out-of-service clients shutdown logic from `ClientConnectionsShutdown` to a separate class `OutOfServiceConnectionsShutdown` which resided in the eureka module.
- Remove `DiscoveryEnabledServer` related logic from `RequestAttempt`'s constructor and moved it to the `newRequestAttempt` method in a new implementation of `NettyOrigin`: `DiscoveryEnabledNettyOrigin` which extends `BasicNettyOrigin`.


asafalima:
@carl-mastrangelo  @artgon I would really appreciate your input about this change.

asafalima:
@carl-mastrangelo @artgon I need your help with pushing it forward

artgon:
Hi @asafalima, this is a reasonable goal but this would break a lot of stuff for us internally. We are too coupled to Eureka and Ribbon in our internal Zuul. 

We do have some future plans to address these coupling problems. Likely in the next major release of Zuul.

mattnelson:
> * Eureka uses Jersey 1.x and JAX-RS 1.x which conflicts with our rest-sever which is based on Jersey 2.x

@asafalima Eureka has jersey2 implementations available.

https://github.com/Netflix/eureka/tree/master/eureka-client-jersey2
https://github.com/Netflix/eureka/tree/master/eureka-core-jersey2

asafalima:
Hi @artgon, 

I saw @carl-mastrangelo indeed noted this as one of the goals for V3 in #771 
Maybe we can make some changes to the work that was already done here to make it fit?

PR #719, opened 2020-02-05
zuul-core: make all HttpSyncEndpoints return SYNC

carl-mastrangelo:
When the BaseZuulFilterRunner runs the endpoints, it always results in a thread hop, even for the ErrorResponse endpoint.  Return the correct type so it's easier to follow through the code.


artgon:
Not sure this is necessary?

PR #722, opened 2020-02-11
Add Cause of SSL Handshake Error from Origin

SYLLL:
When SSL handler reads the message that the handshake failed (for example, because peer certificates are not verified), it populates the through handlers and eventually the failure message goes away. When ProxyEndpoint later tries to talk to origin it usually gets the error of SSLEngine Closed already, which does not display the actual reason why it was closed. This change catches the detailed failure cause and adds it to proxy attempt so client can know what went wrong:

"error":"General OpenSslEngine problem","cause":"java.security.cert.CertificateException: Certificate application name (applicationname) does not match expected name: expectedname","exceptionType":"SSLHandshakeException"

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/OriginResponseReceiver.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/OriginResponseReceiver.java
@@ -194,6 +194,7 @@ public class OriginResponseReceiver extends ChannelDuplexHandler {
                         // Set the specific SSL handshake error if the handlers have already caught them
                         ctx.channel().attr(SSL_HANDSHAKE_UNSUCCESS_FROM_ORIGIN_THROWABLE).set(null);
                         fireWriteError("request headers", cause, ctx);
+                        LOG.debug("SSLException is overridden by SSLHandshakeException caught in handler level. Original SSL exception message: ", future.cause());
                     } else {
                         fireWriteError("request headers", future.cause(), ctx);
                     }
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/OriginResponseReceiver.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/OriginResponseReceiver.java
@@ -34,8 +34,10 @@ import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpResponse;
 import io.netty.handler.codec.http.HttpVersion;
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/niws/RequestAttempt.java
```diff
@@ -325,6 +326,9 @@ else if (t instanceof OutboundException) {
                 error = t.getMessage();
                 exceptionType = t.getClass().getSimpleName();
             }
+            if (t.getCause() != null) {
+                cause = t.getCause().toString();
```
carl-mastrangelo: This should be Throwables.getStackTraceAsString, which correctly expands the causes and suppressed exceptions and stacks.

zuul-core/src/main/java/com/netflix/zuul/netty/server/OriginResponseReceiver.java
```diff
@@ -182,7 +189,14 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
         if (msg instanceof HttpRequestMessage) {
             promise.addListener((future) -> {
                 if (!future.isSuccess()) {
-                    fireWriteError("request headers", future.cause(), ctx);
+                    Throwable cause = ctx.channel().attr(SSL_HANDSHAKE_UNSUCCESS_FROM_ORIGIN_THROWABLE).get();
+                    if (cause != null) {
+                        // Set the specific SSL handshake error if the handlers have already caught them
+                        ctx.channel().attr(SSL_HANDSHAKE_UNSUCCESS_FROM_ORIGIN_THROWABLE).set(null);
+                        fireWriteError("request headers", cause, ctx);
```
carl-mastrangelo: In this case, it would be prudent to log the original `future.cause()` at a debug level, just to make sure it isn't lost.   In the SSL handshake failures, it _usually_ is logged elsewhere, but it isn't guaranteed.   Logging it here would help debugging live instances incase there is additional context in it.

zuul-core/src/main/java/com/netflix/zuul/niws/RequestAttempt.java
artgon: Good point. I think what we really want here is `getMessage()`.

zuul-core/src/main/java/com/netflix/zuul/netty/server/OriginResponseReceiver.java
SYLLL: Sounds good.

PR #723, opened 2020-02-13
zuul-core: wire socket address from ClientChannelManager down to ClientConnectFactory

carl-mastrangelo:
This is needed to ensure IPv6 support, and additionally to pass in LocalAddresses for testing in the future.

This avoids passing a string host around, and forces name resolution to happen earlier in the connection.   The address to connect to is no longer owned by the `Server` object, and instead part of the connection pool.  The Server object now functions as a way to logically reference a backend.

Additionally, several arguments are no longer used and have been removed, like the path, method, and attempt number in multiple places.   

Lastly, blocking DNS reads are still possible, but this change is complicated as it is.   I will make this async in a future PR.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
@@ -442,6 +442,7 @@ public class DefaultClientChannelManager implements ClientChannelManager {
                 stats,
                 instanceInfo,
                 serverAddr,
+                /* serverAddrOverride= */ null,
                 clientConnFactory,
                 pcf,
                 connPoolConfig,
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
@@ -20,6 +20,7 @@ import static com.netflix.client.config.CommonClientConfigKey.NFLoadBalancerClas
 
 import com.google.common.base.Throwables;
 import com.google.common.collect.Sets;
+import com.google.common.net.InetAddresses;
 import com.netflix.appinfo.InstanceInfo;
 import com.netflix.client.config.IClientConfig;
 import com.netflix.loadbalancer.DynamicServerListLoadBalancer;
@@ -43,6 +44,8 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.EventLoop;
 import io.netty.handler.timeout.IdleStateHandler;
 import io.netty.util.concurrent.Promise;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
```diff
@@ -341,12 +341,68 @@ public boolean remove(PooledConnection conn) {
...
+            InetAddress ipAddr = InetAddresses.forString(rawHost);
+            serverAddr = new InetSocketAddress(ipAddr, port);
+        } catch (IllegalArgumentException e1) {
+            LOG.warn("NettyClientConnectionFactory got an unresolved address, addr: {}", rawHost);
+            Counter unresolvedDiscoveryHost = SpectatorUtils.newCounter(
+                    "unresolvedDiscoveryHost",
+                    connPoolConfig.getOriginName() == null ? "unknownOrigin" : connPoolConfig.getOriginName());
+            unresolvedDiscoveryHost.increment();
+            try {
+                serverAddr = new InetSocketAddress(rawHost, port);
```
artgon: Is this second round required because sometimes we may not get an IP above?

zuul-core/src/main/java/com/netflix/zuul/origins/BasicNettyOriginManager.java
```diff
@@ -52,3 +52,4 @@ public BasicNettyOrigin createOrigin(String name, String vip, String uri, boolea
         return new BasicNettyOrigin(name, vip, registry);
     }
 }
+class foo {}
```
artgon: bar?

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
```diff
@@ -373,6 +385,25 @@ public boolean remove(PooledConnection conn) {
...
```
argha-c: If the intention here is to create a unresolved addr, perhaps use `createUnresolved`? Not sure I understand this, after we know the resolution attempt failed.

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/NettyClientConnectionFactory.java
```diff
@@ -26,42 +24,29 @@
...
         if (socketAddress instanceof InetSocketAddress) {
-            InetSocketAddress inetSocketAddress = (InetSocketAddress) socketAddress;
-            // TODO(cmastrangelo): Make this throw an exception after checking the counter isn't implemented, and do
-            //  the name resolution at a higher level.
-            if (inetSocketAddress.isUnresolved()) {
-                LOGGER.warn("NettyClientConnectionFactory got an unresolved address, addr: {}", inetSocketAddress);
-                unresolvedDiscoveryHost.increment();
-            }
+            // This should be checked by the ClientConnectionManager
+            assert !((InetSocketAddress) socketAddress).isUnresolved() : socketAddress;
```
argha-c: This should ideally live in `PerServerConnectionPool::tryMakingNewConnection` ?

zuul-core/src/main/java/com/netflix/zuul/origins/BasicNettyOriginManager.java
argha-c: baz?

carl-mastrangelo: Whoops, I guess that's why I ask for code reviews. :P

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/NettyClientConnectionFactory.java
carl-mastrangelo: It needs to live in CCM because it can return a promise.  PSCP doesn't return a promise, and gets triggered multiple times.    I'd prefer to assert that it is resolved here just before it is actually used, because this method is the one that cares the most about it.   Having the check and the usage very close to each other help understandability.

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
carl-mastrangelo: Correct.   Zuul at Netflix doesn't run into this, but other code users might.   Just preserving any pre-existing behavior.

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
carl-mastrangelo: createUnresolved would allow an unresolved address to propagate down the call chain to NettyClientConnectionFactory, which asserts that should not be done.

As mentioned in the PR description, I plan on rewriting this function to be the point of async name resolution.

carl-mastrangelo:
Made some updates, PTAL

artgon:
Still has class foo?

carl-mastrangelo:
`git add` apparently is needed before `git commit` to have an effect. :/

PR #724, opened 2020-02-19
zuul-core: rethrow bind exceptions with more detail

carl-mastrangelo:


zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -254,7 +256,14 @@ private ChannelFuture setupServerBootstrap(
         serverStatusManager.localStatus(InstanceInfo.InstanceStatus.UP);
 
         // Bind and start to accept incoming connections.
-        return serverBootstrap.bind(listenAddress).sync();
+        ChannelFuture bindFuture = serverBootstrap.bind(listenAddress);
+        try {
+            return bindFuture.sync();
+        } catch (Exception e) {
```
argha-c: Could you catch a more specific exception?

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -254,7 +256,14 @@ private ChannelFuture setupServerBootstrap(
...
+            // sync() sneakily throws a checked Exception, but doesn't declare it. This can happen if there is a bind
+            // failure, which is typically an IOException.  Just chain it and rethrow.
+            throw new RuntimeException("Failed to bind on addr " + listenAddress, e);
```
argha-c: Consider using `BindException` ?

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: I can't because the sync method doesn't actually declare it.   It's a compile error to catch something more specific.

zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
carl-mastrangelo: Alas, I don't want to change the signatures too much, it would be bigger than I had hoped.

PR #731, opened 2020-02-25
zuul-core: Drop malformed requests before passing them up the filter chain

carl-mastrangelo:


zuul-core/src/main/java/com/netflix/netty/common/throttle/RejectionUtils.java
```diff
@@ -49,7 +49,7 @@
 
     // TODO(carl-mastrangelo): add tests for this.
 
-    private static final HttpResponseStatus REJECT_CLOSING_STATUS = new HttpResponseStatus(999, "Closing(Rejection)");
+    public static final HttpResponseStatus REJECT_CLOSING_STATUS = new HttpResponseStatus(999, "Closing(Rejection)");
```
carl-mastrangelo: This needs to be public for the rejection handler to correctly set the http code to -1.   The response status cant be -1 when passed to the the ctor of this class, so I need to check it at the rejection handler.

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -141,6 +140,9 @@ public void channelRead(final ChannelHandlerContext ctx, Object msg) throws Exce
                 zuulRequest.getContext().setShouldSendErrorResponse(true);
             }
 
+            handleExpect100Continue(ctx, clientRequest);
```
carl-mastrangelo: I moved this down below the error handling.   I *think* this is correct, but I may be mistaken.

carl-mastrangelo:
Ideas on how to test this are welcome.

Issue #735, opened 2020-02-29
[Defect] Need ability to support SNI


sandy-adi:
Currently when zuul routes to origins over https, it creates an SSLHandler without setting the SNI. This works for most cases, but for cases where the service may require SNI the SSL handshake fails. Here's the snippet from the [initChannel](https://github.com/Netflix/zuul/blob/ceee0a3f40d043ad57acb86b8e2a5e24ac3ba734/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultOriginChannelInitializer.java#L64) method of [DefaultOriginChannelInitializer](https://github.com/Netflix/zuul/blob/ceee0a3f40d043ad57acb86b8e2a5e24ac3ba734/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultOriginChannelInitializer.java)

```
if (connectionPoolConfig.isSecure()) {
   pipeline.addLast("ssl", sslContext.newHandler(ch.alloc()));
}
```

if this can be updated to something along the lines below, it will help route to origins that need the SNI

```
SslHandler sslHandler = null;
if (connectionPoolConfig.isSecure()) {
    String sni = connectionPoolConfig.getSNI();
    if (sni != null) {
        int securePort = connectionPoolConfig.getSecurePort();
        sslHandler = sslContext.newHandler(ch.alloc(), sni, securePort);
    }else {
        sslHandler = sslContext.newHandler(ch.alloc());
    }
    pipeline.addLast("ssl", sslHandler);
}
```

Please let me know if this looks like the right approach, We've an issue right now because of this and would like to address it as soon as possible.

artgon:
What's the use case for having SNI on the client side?

sandy-adi:
> What's the use case for having SNI on the client side?

oh no the client doesn't need it. The service that is being proxied via zuul needs it. So we need the ability to set the peer host when creating the ssl handler.

artgon:
I'm still not clear on the use case. You want to proxy the SNI bytes from the incoming request to the outbound request? Why does your backend need SNI?

sandy-adi:
The backend [servers are behind a load balancer](https://www.cloudflare.com/learning/ssl/what-is-sni/). This is a common use case where the client has to send in the SNI to the server.

sandy-adi:
@artgon did that make sense?

artgon:
[...]

PR #739, opened 2020-03-05
fix:support extension/customization of netty origin for ssl connections

sandy-adi:
The changes here will enable customization/extension of the Netty components. Another approach to address #735


sandy-adi:
@carl-mastrangelo @artgon could you please take a look at this?

PR #741, opened 2020-03-07
Create an annotation processor which can identify Zuul Filters based on an annotation

carl-mastrangelo:


artgon:
What's the goal of this?

carl-mastrangelo:
The goal is to come up with a compile time list of all filters, and load from that, rather than scanning the classpath for filters.    This should speed up boot time of Zuul a lot.

PR #744, opened 2020-03-13
all: update spectator api

carl-mastrangelo:


carl-mastrangelo:
This is causing skew between our internal and external versions.

PR #758, opened 2020-03-30
zuul-core: remove reference to Servo Counter

carl-mastrangelo:


carl-mastrangelo:
The goal of this PR is to start chipping away at our Servo dependency.  It is deprecated in favor of Spectator, which is actively maintained.

PR #762, opened 2020-04-02
zuul-core: user sorted set on FilterLoader, and accept set on StaticFilterLoader

carl-mastrangelo:


zuul-core/src/test/java/com/netflix/zuul/StaticFilterLoaderTest.java
```diff
@@ -16,45 +16,47 @@
...
     @Test
     public void getFilterByNameAndType() {
         StaticFilterLoader filterLoader =
-                new StaticFilterLoader(factory, Arrays.asList(DummyFilter2.class, DummyFilter1.class));
+                new StaticFilterLoader(factory, ImmutableSet.of(DummyFilter2.class, DummyFilter1.class));
 
         ZuulFilter<?, ?> filter = filterLoader.getFilterByNameAndType("Robin", FilterType.INBOUND);
 
-        assertEquals(DummyFilter2.class, filter.getClass());
+        Truth.assertThat(filter).isInstanceOf(DummyFilter2.class);
```
argha-c, line 59: Why use a different assertion style here?

zuul-core/src/main/java/com/netflix/zuul/FilterLoader.java
```diff
@@ -48,9 +50,12 @@
...
     /**
-     * Returns a list of filters by the filterType specified.
+     * Returns a sorted set of filters by the filterType specified.
      */
-    List<ZuulFilter<?, ?>> getFiltersByType(FilterType filterType);
+    SortedSet<ZuulFilter<?, ?>> getFiltersByType(FilterType filterType);
 
     ZuulFilter<?, ?> getFilterByNameAndType(String name, FilterType type);
+
+    Comparator<ZuulFilter<?, ?>> FILTER_COMPARATOR =
```
argha-c, line 59: Nice!

zuul-core/src/test/java/com/netflix/zuul/StaticFilterLoaderTest.java
carl-mastrangelo: It has some very nice list/set/string assertions.   I had added this a few months ago but had to revert it for an unrelated reason.   It's a very nice library to use, and I was planning on replacing some other usages too.

PR #775, opened 2020-04-12
zuul-core: HashMap's key override equals and hashCode method

ctlove0523:
`ChannelConfigKey` used as HashMap's key in `ChannelConfig`,so `ChannelConfigKey`  override equals and hashCode method is a good way to prevent potential memory leak.


carl-mastrangelo:
Hi @ctlove0523 I think the intent with these keys was for them to be singletons, and compared based on identity.  Can you describe your use case more?

ctlove0523:
@carl-mastrangelo oh I understand what you means,I will check my codes.

PR #776, opened 2020-04-12
zuul-core,zuul-guice: factor guice dependency to separate module

carl-mastrangelo:


carl-mastrangelo:
Not entirely sure on the direction here, but maybe we can also make a zuul-spring version sometime in the future.

artgon:
I think the general direction makes sense. It would be nice to be able to sub in guice/spring/etc as needed.

zuul-core/src/main/java/com/netflix/zuul/netty/server/BaseServerStartup.java
```diff
@@ -102,7 +100,7 @@ public Server server()
         return server;
     }
 
-    @PostConstruct
+    @Inject
```
asafalima, line 103: Why did you replace all `@PostConstruct` with `@Inject`?
This annotation is part of JSR250, guice doesn't have builtin support for it but there are number of libraries which solve that problem like governator and guice-bootstrap.
Spring have builtin support of it.

I think it's a bit unclear the usage of `@Inject` for initializing classes.

carl-mastrangelo: Method injection runs after the Constructor has run, it's similar to PostConstruct and recommended by Guice.  I don't think lifecycle calls should be part of the injection story.

argha-c: +1 on  vanilla injection without lifecycle support. This does not in any way block the support for custom lifecycle hooks IMHO.

asafm: @carl-mastrangelo I searched in Method Injection documentation both in Guice and in JSR-330 and couldn't find any reference for recommendations. All I saw was a way to inject dependencies via methods. To me it also strikes as counter intuitive and "hackish". You're not really injecting anything. PostConstuct which is part of the JSR was created exactly for that purpose.

carl-mastrangelo: Ah, you're right, I found it on the Dagger GitHub:

https://github.com/google/dagger/issues/455#issuecomment-245098467

For reference: the people responding there are on the Java core libraries at Google, and maintainers of Guice.  

@asafm If you don't like using the Method injection style, I'd like to understand why?  Having post construction logic is itself an anti pattern, so I don't believe this makes things worse.   You call it hackish, but look at it more closely:

* The method name is called "init", indicating it runs after objection construction
* Tests that run this code work correctly. It can be verified experimentally that it's called in the correct order.
* The docs for `@Inject` indicate that Method injection happens after all Construction and Field injection.

So, it's well documented what's going on, and tests show that the documentation is being followed.   I think that makes it okay for a feature that arguably shouldn't have been used in the first place.  I'd be happy to remove it if the logic was factored in such a way that post construction isn't necessary.

asafm: I'll try to explain my view better, with examples. 

`@PostConstruct` isn't really an anti-pattern (Even Google doesn't yield this, as opposed to other super debatable topics). Let's start backwards to explain why it's important. 

Let's say you have a bean called Server. It's probably have a `start()` method and `shutdown()`. Same goes for any other object - for example one that have a thread pool, so in the `start()` it submits a bunch of tasks to it, and on `shutdown()` it awaits for task to gracefully stop and then stop any other tasks from being submitted and "kills" threads in thread pool. 

Graceful stop is something you want both in production and in testing:
* In production - When your container gets killed (let's say K8s moves it to another pod), you want to gracefully stop any existing activity for a given timeout. Suppose the bean I described has an internal queue and thread dumping it to Elasticsearch. On graceful shutdown, you wait for the queue to be written to ES and then stop. So graceful stops is essentially good for avoiding data loss and prevents inaccuracies in your data. Of course there is the machine killed scenario, but it's something you can't control and happens less often than deployments (for example).
* In tests - If you run integration test (e.g. unit of test is your microservice), you would start and stop your Main class (function) running your microservice multiple times (each time with different configuration), using same JVM. So naturally you would want a graceful shutdown of the beans declared in your DI container, to avoid leaving tasks you don't want to run in parallel, thread leaks, etc. 

Given you agree to the above mentioned explanations (also same thread you gave me, see further comments: https://github.com/google/dagger/issues/455#issuecomment-313666075), thus it's quite clear you would want `@PreDestroy` support in your DI.

Now, given you have a `shutdown()` method for your class, a more readable code would  have `start()` method to accompany that, so you would associate what's being started to what's being shutdown. Hence you want support for` @PostConstruct`.

Of course you can settle for using `@Inject` instead of `@PostConstruct` on that `start()` method, but I believe it's not as obvious and readable relative to `@PostConstruct`. The latter says clearly: It's a Post Construction method, while the former says: Inject - this method will be used to inject - but when you look at its arguments, nothing is really injected. Another problem is discoverability: People using Zuul, would want to add their own objects, or just read/modify existing code. It will be quite a surprise for them to learn that a `start()` method should be annotated with `@Inject` and not `@PostConstruct`. Of course you can tell them to place that code in the constructor, but then you are harming their code readability, and also Zuul code readability. 

`@PostConstruct` has become a familiar standard. It's defined in the JSR, and also supported in Spring, Guice, Weld (Most popular ones I know of). Dagger took their own path, much like Kodein did: off the beaten path. 

Regarding the last bit of: "It's well documented". I personally prefer code which self explains it self vs reading the Javadoc to "get" it. Reading @Inject javadoc won't really explain why a argument-less method injection is in fact a initialization method. Most likely it would confuse at first the reader. Then he would have the "a-ha" moment and then he will have to get used to it and remember. Repeat per set of eyes reading this. 

I hope I managed to explain:
1. Why life-cycle support is needed: `@PostConstruct` and `@PreDestroy` and why it should be used.
2. Why `@PostConstruct` isn't equal to argument-less method injection
3. How having both will improve code readability and familiarity across developer using Zuul and contributing to it.

carl-mastrangelo: I read your response, I think the point of disagreement is that DI should be used for lifecycle management.   Calling start() and shutdown() is something your code, which knows the lifecycles of your objects, should be doing.  

Also, bringing this thread back, is this affecting your code?   If this is causing some damage I would be able to work with you on fixing it.

PR #778, opened 2020-04-15
zuul-core: provide a Netty Promise adapter

carl-mastrangelo:


artgon:
LGTM, but this is a pretty fundamental change. Should we have a bigger discussion about this? Also, can we add some tests to verify the behavior?

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -233,36 +232,34 @@ protected final O filter(final ZuulFilter<I, O> filter, final I inMesg) {
...
-                        .observeOn(Schedulers.from(getChannelHandlerContext(inMesg).executor()))
-                        .doOnUnsubscribe(resumer::decrementConcurrency)
-                        .subscribe(resumer);
+                filter.filterAsync(promise, inMesg);
             } finally {
-              PerfMark.stopTask(filter.filterName(), "applyAsync");
+                PerfMark.stopTask(filter.filterName(), "filterAsync");
             }
 
+            if (promise.isSuccess()) {
```
argha-c, line 247: What's the benefit of handling success here, as opposed to always using the listener callback, regardless of result?

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -396,116 +392,112 @@ protected void resumeInBindingContext(final O zuulMesg, final String filterName)
         }
     }
 
-    private final class FilterChainResumer implements Observer<O> {
-        private final I inMesg;
+    private final class FilterChainResumer implements GenericFutureListener<Future<O>> {
```
argha-c, line 395: `FutureListener<0>`?

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
carl-mastrangelo: Returning early here allows calling the next filter immediately, rather than yielding the thread and handling another request.  It's an optimization.

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
carl-mastrangelo: Done.

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
argha-c: Thought so. Thanks for adding the comments.

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -380,97 +397,115 @@ protected void resumeInBindingContext(final O zuulMesg, final String filterName)
...
             }
-            catch (Exception e) {
-                decrementConcurrency();
-                handleException(inMesg, filter.filterName(), e);
+
+        }
+
+        private void handleFailure(Throwable t) {
+            try {
+                recordFilterCompletion(FAILED, filter, startTime, inMsg, snapshot);
```
SYLLL, line 440: NIT: why not putting recordFilterCompletion after resumeInBindingContext

zuul-core/src/main/java/com/netflix/zuul/filters/FilterHelper.java
```diff
@@ -0,0 +1,62 @@
...
+    static <O> void observerToPromise(Promise<? super O> promise, Observable<? extends O> observable) {
+        // TODO(carl-mastrangelo): add tracing links for this adapter.
+        observable.subscribe(new Observer<O>() {
+            @Override
+            public void onCompleted() {}
+
+            @Override
+            public void onError(Throwable t) {
+                if (!promise.tryFailure(t)) {
+                    if (!promise.isSuccess()) {
```
SYLLL, line 38: I see similarities between line 38-43 and line 51-55, maybe you could consider combining them

Issue #780, opened 2020-04-15
[Defect] Zuul-Core does not use updated IP address for a given DNS without app restart


hermanyglauber:
Scenario: 
1. There is a service registered in eureka with 'service.company.aws' DNS enpoint. That DNS resolves to IP: 192.168.0.5.
2. A request is made to the service endpoint via Zuul creating the pool with DNS/IP as serverAddr.
3. The service.company.aws's IP changes to 192.168.0.17
4. A request is made to the service via Zuul.

Expected Result:
The new DNS lookup should fetch the new IP (192.168.0.17) and set it in the connection pool config

Actual Result:
Zuul is using the old IP (192.168.0.5) address stored in perServerConnectionPool resulting in a 502 Bad Gateway error.

https://github.com/Netflix/zuul/blob/master/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java#L399

The DNS lookup seems to be always happening and it is being assigned to the finalServerAddr var but the pool is not updated with it. 

```
           try {
                serverAddr = new InetSocketAddress(rawHost, port);
            } catch (RuntimeException e2) {
                e1.addSuppressed(e2);
                throw e1;
            }
        }
        final InetSocketAddress finalServerAddr = serverAddr;

        selectedServer.set(chosenServer);
        // Now get the connection-pool for this server.
        IConnectionPool pool = perServerPools.computeIfAbsent(chosenServer, s -> {
            // Get the stats from LB for this server.
            LoadBalancerStats lbStats = loadBalancer.getLoadBalancerStats();
            ServerStats stats = lbStats.getSingleServerStat(chosenServer);

            final ClientChannelManager clientChannelMgr = this;
            PooledConnectionFactory pcf = createPooledConnectionFactory(chosenServer, instanceInfo, stats, clientChannelMgr, closeConnCounter, closeWrtBusyConnCounter);

            // Create a new pool for this server.
            return createConnectionPool(chosenServer, stats, instanceInfo, finalServerAddr, clientConnFactory, pcf, connPoolConfig,
                    clientConfig, createNewConnCounter, createConnSucceededCounter, createConnFailedCounter,
                    requestConnCounter, reuseConnCounter, connTakenFromPoolIsNotOpen, maxConnsPerHostExceededCounter,
                    connEstablishTimer, connsInPool, connsInUse);
        });

        return pool.acquire(eventLoop, passport, selectedHostAddr);
```

[...]

PR #781, opened 2020-04-16
refactor: always set the server address on service connection pool

hermanyglauber:
Logic to avoid DNS/IP address caching on perServerConnectionPool that causes 502 Bad Gateway error connecting to the Origin endpoint. Issue described on https://github.com/Netflix/zuul/issues/780

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
@@ -407,7 +407,7 @@ public class DefaultClientChannelManager implements ClientChannelManager {
         selectedServer.set(chosenServer);
 
         // Now get the connection-pool for this server.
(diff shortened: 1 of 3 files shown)


carl-mastrangelo:
I think this may not be correct.   The per-server connection pool should probably be torn down and recreated when the address changes.

hermanyglauber:
I thought about that possibility as well but didn't test it. I will refactor it and test the effect.

carl-mastrangelo:
Just FYI, this may be a bigger piece of work than you want to bite off.  I think the pool needs to be closed (witch existing requests being allowed to finish).   This will be tough to test.

hermanyglauber:
FYI, those changes have been deployed in 6 different environments in a custom zuul-core version. Performance test has been done, there was no degradation in the resources utilization, neither in latency and neither in error rate, in fact, the number of request with 5xx status decreased and so did retries. 

Most likely the in progress connections are going to fail as the backend will not be connectable via old IP, do you agree?

PR #816, opened 2020-05-27
Add blacklist for allowing XFF headers

artgon:
A simple convenience property for disallowing XFF from certain hosts.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Netflix, Inc.
+ * Copyright 2020 Netflix, Inc.
  *
  *      Licensed under the Apache License, Version 2.0 (the "License");
  *      you may not use this file except in compliance with the License.
--- a/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
@@ -118,7 +118,7 @@ public class StripUntrustedProxyHeadersHandler extends ChannelInboundHandlerAdap
     @VisibleForTesting
     void checkBlacklist(HttpRequest req, List<String> blacklist) {
         // blacklist headers from
-        if (blacklist.contains(req.headers().get(HttpHeaderNames.HOST).toLowerCase())) {
+        if (blacklist.stream().anyMatch(h -> h.equalsIgnoreCase(req.headers().get(HttpHeaderNames.HOST)))) {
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
```diff
@@ -96,11 +106,20 @@ private boolean connectionIsUsingMutualSSLWithAuthEnforced(Channel ch)
...
         HttpHeaders headers = req.headers();
         for (AsciiString headerName : HEADERS_TO_STRIP) {
             headers.remove(headerName);
         }
     }
+
+    @VisibleForTesting
+    void checkBlacklist(HttpRequest req, List<String> blacklist) {
+        // blacklist headers from
+        if (blacklist.contains(req.headers().get(HttpHeaderNames.HOST).toLowerCase())) {
```
carl-mastrangelo, line 121: There should be a `normalized()` method you can call that avoids the lowercase string allocation.

zuul-core/src/test/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandlerTest.java
```diff
@@ -0,0 +1,137 @@
...
+        when(headers.get(eq(HttpHeaderNames.HOST))).thenReturn("netflix.com");
+    }
+
+    @Test
+    public void allow_never() throws Exception {
+        StripUntrustedProxyHeadersHandler stripHandler = getHandler(AllowWhen.NEVER);
+
+        stripHandler.channelRead(channelHandlerContext, msg);
+
+        verify(stripHandler).stripXFFHeaders(any());
```
carl-mastrangelo, line 82: wouldn't it be more accurate to test that the headers are removed?

zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
artgon: Unfortunately it's a Netty header object and not ours. Do you know if they handle casing on gets? If they do, we can just remove the line.

zuul-core/src/test/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandlerTest.java
artgon: In a sense yes, but maybe not for this specific unit. I could another test that validates that we're stripping headers properly.

zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
carl-mastrangelo: Hmm, I guess not.  consider using equalsIgnoreCase to avoid the alloc.

PR #830, opened 2020-06-25
zuul-core: Add utility for case insensitive lookup of a query param

argha-c:


zuul-core/src/test/java/com/netflix/zuul/message/http/HttpQueryParamsTest.java
```diff
@@ -17,8 +17,12 @@
 package com.netflix.zuul.message.http;
 
 
+import static com.google.common.truth.Truth.assertThat;
```
carl-mastrangelo, line 20: unused import?

argha-c: Fixed.

PR #841, opened 2020-07-06
all: bump spectator and jackson

carl-mastrangelo:


carl-mastrangelo:
test is a flake, submitting

PR #851, opened 2020-07-15
Sanitize header fields and values

artgon:
This will prevent headers set in filters/handlers from inadvertently being invalid or broken.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
@@ -503,7 +503,7 @@ public final class Headers {
     private String sanitizeField(String value) {
         if (value != null) {
             int l = value.length();
-            StringBuilder clean = new StringBuilder();
+            StringBuilder clean = new StringBuilder(l);
             for (int i = 0; i < l; i++) {
                 char c = value.charAt(i);
                 // ASCII non-control characters, per RFC 7230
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -499,4 +499,20 @@ private void truncate(int i) {
             values.remove(k);
         }
     }
+
+    private String sanitizeField(String value) {
```
carl-mastrangelo, line 503: sanitization is an attack vector.   I feel that if the value is bad, the header should be thrown away.  (or the respond with a 400).

artgon: I'm open to throwing an exception in that method instead of sanitizing. Curious what the attack vector is here?

carl-mastrangelo: Mutating the response I think it incorrect.   What if by stripping chars, we end up making something that would previously be blocked?   We need to only produce good data.   I think throwing an exception is reasonable.

Also, if we get data in from netty, it may be "invalid", so this would only protect stuff produced by our filters.  I think the validation would need to happen down at the netty layer.

artgon: Yeah, I think you're right. I'll make it throw an exception.

On the second point, I believe the Netty HTTP codec will protect us from incoming parsing errors but this will protect us from the filters settings bad headers.

PR #854, opened 2020-07-15
use separate API for validation not to break existing functionality

artgon:


artgon:
This is effectively a rollback of the previous PR.

PR #865, opened 2020-07-30
Respect Proxy Protocol value of original client IP address

ldayton:
Use proxy protocol port for Zuul `HttpRequestMessage.getOriginalPort()` if available

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/context/CommonContextKeys.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/context/CommonContextKeys.java
@@ -43,6 +43,7 @@ public class CommonContextKeys {
     public static final String USE_FULL_VIP_NAME = "use_full_vip_name";
     public static final String ACTUAL_VIP = "origin_vip_actual";
     public static final String ORIGIN_VIP_SECURE = "origin_vip_secure";
+    public static final String REMOTE_PORT = "remote_port";
 
     public static final String SSL_HANDSHAKE_INFO = "ssl_handshake_info";
 
--- a/zuul-core/src/main/java/com/netflix/zuul/context/CommonContextKeys.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/context/CommonContextKeys.java
@@ -43,7 +43,12 @@ public class CommonContextKeys {
     public static final String USE_FULL_VIP_NAME = "use_full_vip_name";
     public static final String ACTUAL_VIP = "origin_vip_actual";
     public static final String ORIGIN_VIP_SECURE = "origin_vip_secure";
-    public static final String REMOTE_PORT = "remote_port";
+
+    /**
+     * The original client port reported to Zuul by a proxy running Proxy Protocol.
+     * Will only be set if both Zuul and the connected proxy are both using set to use Proxy Protocol.
+     */
+    public static final String PROXY_PROTOCOL_PORT = "proxy_protocol";
 
     public static final String SSL_HANDSHAKE_INFO = "ssl_handshake_info";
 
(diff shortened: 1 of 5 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -522,14 +522,17 @@ public String getOriginalProtocol()
...
+            return getOriginalPort(getContext(), getHeaders(), getPort());
         } catch (URISyntaxException e) {
             throw new IllegalArgumentException(e);
         }
     }
 
     @VisibleForTesting
-    static int getOriginalPort(Headers headers, int serverPort) throws URISyntaxException {
+    static int getOriginalPort(SessionContext context, Headers headers, int serverPort) throws URISyntaxException {
+        if (context.containsKey(CommonContextKeys.REMOTE_PORT)) {
```
artgon, line 533: Not sure if this is intended, but `REMOTE_PORT` will take precedence over the XFF header. This decision may need to be made based on whether we actually have PPv2 data or not, or even whether we prefer it.

ldayton: Thanks for catching that. I didn't realize `ATTR_REMOTE_ADDR` was available if PPv2 wasn't available. Looking now I see that it's set in `SourceAddressChannelHandler.channelActive()`. Will update.

As for precedence over XFF header, in the case of a chain of proxies either XFF or PPv2 source IP could be wrong by the time it gets to the current proxy. Not sure how precedence should be handled.

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -267,6 +269,10 @@ private HttpRequestMessage buildZuulHttpRequest(
 
         // This is the only way I found to get the port of the request with netty...
         final int port = channel.attr(SourceAddressChannelHandler.ATTR_SERVER_LOCAL_PORT).get();
+        final SocketAddress remoteAddress = channel.attr(SourceAddressChannelHandler.ATTR_REMOTE_ADDR).get();
+        if (remoteAddress instanceof InetSocketAddress) {
+            context.set(CommonContextKeys.REMOTE_PORT, ((InetSocketAddress) remoteAddress).getPort());
```
argha-c, line 274: I believe what you're trying to do exists in the form of `HttpRequestMessageImpl#clientRemoteAddress`. 
The issue is likely that I set that value to `ATTR_LOCAL_ADDR` instead of `ATTR_REMOTE_ADDR`, as you correctly do here.
See https://github.com/Netflix/zuul/blob/d8838c5131bf1ea58cd141bd2a8cd0ff8ea1ab95/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java#L271
I'd probably try updating that and update the tests, which might fix the issue.

zuul-core/src/main/java/com/netflix/zuul/context/CommonContextKeys.java
```diff
@@ -43,7 +43,12 @@
     public static final String USE_FULL_VIP_NAME = "use_full_vip_name";
     public static final String ACTUAL_VIP = "origin_vip_actual";
     public static final String ORIGIN_VIP_SECURE = "origin_vip_secure";
-    public static final String REMOTE_PORT = "remote_port";
+
+    /**
+     * The original client port reported to Zuul by a proxy running Proxy Protocol.
+     * Will only be set if both Zuul and the connected proxy are both using set to use Proxy Protocol.
+     */
+    public static final String PROXY_PROTOCOL_PORT = "proxy_protocol";
```
artgon, line 51: I think the value should be the same as the variable name.

ldayton: thanks for catching that, updated

PR #869, opened 2020-08-04
Fix x-forwarded-port

ldayton:
Fix bug from previous PR where x-forwarded-for was using PPv2 source address instead of destination


zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -271,7 +271,7 @@ private HttpRequestMessage buildZuulHttpRequest(
         final int port = channel.attr(SourceAddressChannelHandler.ATTR_SERVER_LOCAL_PORT).get();
         final HAProxyMessage haProxyMessage = channel.attr(HAProxyMessageChannelHandler.ATTR_HAPROXY_MESSAGE).get();
         if (haProxyMessage != null) {
-            context.set(CommonContextKeys.PROXY_PROTOCOL_PORT, haProxyMessage.sourcePort());
+            context.set(CommonContextKeys.PROXY_PROTOCOL_PORT, haProxyMessage.destinationPort());
```
argha-c, line 274: As discussed, this should have been set here -  https://github.com/Netflix/zuul/blob/master/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/HAProxyMessageChannelHandler.java#L66-L77

We might need to debug deeper if this change doesn't work help.

PR #871, opened 2020-08-06
Add a basic test for HAPM channel handler

argha-c:


argha-c:
>other than the new proxy protocol attribute, but that line could have been added to the existing tests
Those tests should've been part of the initial change, yes.

The `ElbProxyProtocolChannelHandler` essentially is only responsible for adding the decoder in the pipeline, based on the value of the proxy protocol attribute.

The core of the decoding and statefulness happens in `HAProxyMessageChannelHandler`, so it'd make sense to unit test that logic in it's own test class.

PR #872, opened 2020-08-09
New API to modify header values only if valid

kyagna:
New APIs to set and add headers only if their name and values are valid.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
@@ -667,8 +667,7 @@ public final class Headers {
      * Returns the index of first invalid character. Returns {@link #ABSENT} if absent.
      */
     private static int findInvalid(String value) {
-        int l = value.length();
-        for (int i = 0; i < l; i++) {
+        for (int i = 0; i < value.length(); i++) {
             char c = value.charAt(i);
             // ASCII non-control characters, per RFC 7230 but slightly more lenient
             if (c < 31 || c == 127) {
--- a/zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
@@ -196,6 +196,30 @@ public final class Headers {
     }
 
     /**
+     * Replace any/all entries with this key, with this single entry if the key and entry are valid.
+     *
+     * If value is {@code null}, then not added, but any existing header of same name is removed.
+     */
+    public void setIfValid(HeaderName headerName, String value) {
+        String normalName = requireNonNull(headerName, "headerName").getNormalised();
+        if (isValid(headerName.getName()) && isValid(normalName) && isValid(value)) {
+            setNormal(headerName.getName(), normalName, value);
+        }
+    }
+
+    /**
+     * Replace any/all entries with this key, with this single entry if the key and entry are valid.
+     *
+     * If value is {@code null}, then not added, but any existing header of same name is removed.
+     */
+    public void setIfValid(String headerName, @Nullable String value) {
+        String normalName = HeaderName.normalize(requireNonNull(headerName, "headerName"));
+        if (isValid(headerName) && isValid(normalName) && isValid(value)) {
+            setNormal(headerName, normalName, value);
+        }
+    }
+
+    /**
      * Replace any/all entries with this key, with this single entry and validate.
      *
      * If value is {@code null}, then not added, but any existing header of same name is removed.
@@ -290,6 +314,36 @@ public final class Headers {
     }
 
     /**
+     * Validates and adds the name and value to the headers, except if the name is already present.  Unlike
+     * {@link #set(String, String)}, this method does not accept a {@code null} value.
+     *
+     * @return if the value was successfully added.
+     */
+    public boolean setIfAbsentAndValid(String headerName, String value) {
+        requireNonNull(value, "value");
+        String normalName = HeaderName.normalize(requireNonNull(headerName, "headerName"));
+        if (isValid(headerName) && isValid(normalName) && isValid((value))) {
+            return setIfAbsentNormal(headerName, normalName, value);
+        }
+        return false;
+    }
+
+    /**
+     * Validates and adds the name and value to the headers, except if the name is already present.  Unlike
+     * {@link #set(HeaderName, String)}, this method does not accept a {@code null} value.
+     *
+     * @return if the value was successfully added.
+     */
+    public boolean setIfAbsentAndValid(HeaderName headerName, String value) {
+        requireNonNull(value, "value");
+        String normalName = requireNonNull(headerName, "headerName").getNormalised();
+        if (isValid(headerName.getName()) && isValid(normalName) && isValid((value))) {
+            return setIfAbsentNormal(headerName.getName(), normalName, value);
+        }
+        return false;
+    }
+
+    /**
      * Adds the name and value to the headers.
      */
     public void add(String headerName, String value) {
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +624,53 @@ private void truncate(int i) {
...
+     */
+    private boolean isValid(String value) {
+        if (value != null) {
+            int pos = findInvalid(value);
+            if (pos != ABSENT) {
+                Spectator.globalRegistry().counter("zuul.header.invalid.char").increment();
+                return false;
+            }
+        }
+        return true;
```
samukce, line 638: You could use Guard Clause this way we could have some code readability improvement here. 

```suggestion
        if (value == null) {
            return true;
        }
        if (findInvalid(value) == ABSENT) {
                return true;
        }
        Spectator.globalRegistry().counter("zuul.header.invalid.char").increment();
        return false;
```

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +624,53 @@ private void truncate(int i) {
...
                 if (c < 31 || c == 127) {
-                    Spectator.globalRegistry().counter("zuul.header.invalid.char").increment();
-                    throw new ZuulException("Invalid header field: char " + (int) c + " in string " + value
-                            + " does not comply with RFC 7230");
+                    return i;
                 }
             }
         }
-        return value;
+        return -1;
```
samukce, line 674:
```suggestion
        return ABSENT;
```

zuul-core/src/test/java/com/netflix/zuul/message/HeadersTest.java
```diff
@@ -202,6 +202,30 @@ public void setNullIsEmtpy_headerName() {
...
+        headers.setIfValid("X-Valid-K3", "abc\n-xy\nz");
+        headers.setIfValid("X-Valid-K4", "abc-abc");
+
+        Truth.assertThat(headers.getAll("CookiE")).isEmpty();
+        Truth.assertThat(headers.getAll("X-Valid-K1")).containsExactly("abc-xyz");
+        Truth.assertThat(headers.getAll("X-Valid-K2")).containsExactly("def-xyz");
+        Truth.assertThat(headers.getAll("X-Valid-K3")).containsExactly("xyz-xyz");
+        Truth.assertThat(headers.getAll("X-Valid-K4")).containsExactly("abc-abc");
+        Truth.assertThat(headers.size()).isEqualTo(5);
+    }
```
samukce, line 227: Do you think that is possible to break this test in multiple small tests/behaviors?

kyagna: Done

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -195,6 +195,30 @@ public void setAndValidate(String headerName, @Nullable String value) {
...
     }
 
+    /**
+     * Replace any/all entries with this key, with this single entry if the key and entry are valid.
+     *
+     * If value is {@code null}, then not added, but any existing header of same name is removed.
+     */
+    public void setIfValid(HeaderName headerName, String value) {
+        String normalName = requireNonNull(headerName, "headerName").getNormalised();
+        if (isValid(headerName.getName()) && isValid(normalName) && isValid(value)) {
```
carl-mastrangelo, line 205: I think you should check if it is valid first, and then normalize.   I think it is true that if the non-normal name is valid, the normal one must be as well.  Thus, I dont think you need to validate the normal name.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -195,6 +195,30 @@ public void setAndValidate(String headerName, @Nullable String value) {
...
+    }
+
+    /**
+     * Replace any/all entries with this key, with this single entry if the key and entry are valid.
+     *
+     * If value is {@code null}, then not added, but any existing header of same name is removed.
+     */
+    public void setIfValid(String headerName, @Nullable String value) {
+        String normalName = HeaderName.normalize(requireNonNull(headerName, "headerName"));
+        if (isValid(headerName) && isValid(normalName) && isValid(value)) {
```
carl-mastrangelo, line 217: nit: the validity checks for names and values are different I thought?   It's important that validity is well defined at the creation of this new API method, because it will become hard to change it later.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +624,50 @@ private void truncate(int i) {
         }
     }
 
+    /**
+     * Checks if the given value is compliant with our RFC 7230 based check
+     */
+    private boolean isValid(String value) {
```
carl-mastrangelo, line 630: This should be a static method.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +624,50 @@ private void truncate(int i) {
...
+        if (value == null || findInvalid(value) == ABSENT) {
```
carl-mastrangelo, line 631: I dont think null values should ever reach here.  All public methods check their params before using them, so it shouldn't be necessary to double check here.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +624,50 @@ private void truncate(int i) {
...
+            return true;
+        }
+        Spectator.globalRegistry().counter("zuul.header.invalid.char").increment();
```
carl-mastrangelo, line 634: You can store a reference to this counter in a static variable to avoid recalculating it each time.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +624,50 @@ private void truncate(int i) {
...
+    }
+
+    /**
+     * Validated the input value based on RFC 7230 but more lenient.
+     * Currently, only ASCII control characters are considered invalid.
+     *
+     * Returns the index of first invalid character. Returns {@link #ABSENT} if absent.
+     */
+    private int findInvalid(String value) {
         if (value != null) {
```
carl-mastrangelo, line 661: Again, this should not be checking for null

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +624,50 @@ private void truncate(int i) {
...
+
+    /**
+     * Checks if the input value is compliant with our RFC 7230 based check
+     * Returns input value if valid, raises ZuulException otherwise
+     */
     private String validateField(String value) {
+        if (value != null) {
+            int pos = findInvalid(value);
+            if (pos != ABSENT) {
+                Spectator.globalRegistry().counter("zuul.header.invalid.char").increment();
```
carl-mastrangelo, line 646: Same comment

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
kyagna: Some public methods (like set API and its variants) allow for null as header values. This was the reason for having the null check for isValid and validateField.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
kyagna: Ideally header name validation should be stricter, but want to make sure that doing that does not break any existing behavior. Keeping the validation of header names relaxed in this PR for now.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
carl-mastrangelo: Then it needs to be specified on this method as `@Nullable`

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
```diff
@@ -548,19 +633,48 @@ private void truncate(int i) {
...
     }
+
+    /**
+     * Validated the input value based on RFC 7230 but more lenient.
+     * Currently, only ASCII control characters are considered invalid.
+     *
+     * Returns the index of first invalid character. Returns {@link #ABSENT} if absent.
+     */
+    private static int findInvalid(String value) {
+        int l = value.length();
```
carl-mastrangelo, line 670: no need to extract this out, java will realize the size is a constant in the loop.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
kyagna: Makes sense. Changed it.

zuul-core/src/main/java/com/netflix/zuul/message/Headers.java
kyagna: done

PR #878, opened 2020-08-19
zuul-core: add a conneciton timer class

carl-mastrangelo:
I picked to make a new Attrs class because the channel one doesn't really allow grouping.   It also forces thread safety when we dont need it, and makes keys identical based on name.   These have proved problematic in the past, so I made a new one.

The idea of the conn timer is to store it on the channel itself, along with the set of dimensions to break down stats by.  The attrs map is dynamic, and allows updating it as new metrics come in.   I plan on using these dimensions for other connection stats in the future.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
@@ -23,6 +23,10 @@ import com.netflix.netty.common.CategorizedThreadFactory;
 import com.netflix.netty.common.LeastConnsEventLoopChooserFactory;
 import com.netflix.netty.common.metrics.EventLoopGroupMetrics;
 import com.netflix.netty.common.status.ServerStatusManager;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spectator.api.Spectator;
+import com.netflix.zuul.Attrs;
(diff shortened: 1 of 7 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -442,18 +450,20 @@ synchronized private void stop()
...
+    private final class NewConnHandler extends ChannelInboundHandlerAdapter {
 
         @Override
         public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
             Long now = System.nanoTime();
-            Map<String, Long> timings = new LinkedHashMap<>();
-            timings.put(CONN_ACCEPT, now);
             final Channel child = (Channel) msg;
-            child.attr(CONN_TIMING).set(timings);
+            child.attr(CONN_DIMENSIONS).set(Attrs.newInstance());
```
artgon, line 464: How are these dimensions going to be set?

carl-mastrangelo: I was planning on setting them in the handlers as the different stages of negotiation happen.  Things like ipv4/6, http/1 or 2.   Connection names like the target, or authority for outbound ssl.   The channel is available to the handlers, so this should be acccessible.

PR #891, opened 2020-09-11
core: rename app name extractor and make it allocate less memory

carl-mastrangelo:


troshko111 (commented):
LGTM, making this `final` is potentially a breaking change though.

PR #896, opened 2020-09-24
Modify the type of the Chosen Host address to be an InetAddress.

carl-mastrangelo:
This will allow clearer understanding of what this type can be used as in other parts of the Zuul Code base.   The long term vision is to modify this to not be an InetAddress, but a `ExtendedSocketAddress` which includes the other metaa information about the addr.    The change as is is more to make sure the type is consitent with the current semantics, rather than changing the semantics.

`ZUUL_ORIGIN_ATTEMPT_IPADDR_MAP_KEY` will eventually be deleted, as it only seems to be used for event context logging.    That too will occur in a followup change.


carl-mastrangelo:
Note, I am intentionally making this PR a smaller one and not including all the changes yet to make it easier to review.   This will not be the final state.

artgon (approved):

PR #903, opened 2020-10-06
core: add attrs to request message

carl-mastrangelo:
This is an example of adding Attrs (as an alternative to SessionContext).  Attrs are type safe, and make it easy to find out where a particular Key is used.  Attrs is mutable as there are several places where it its mutated.    Since we don't have the error prone checks to ensure return values are used, it would be risky to make Attrs immutable.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
@@ -19,6 +19,7 @@ import com.google.common.base.Charsets;
 import com.google.common.base.Strings;
 import com.netflix.config.DynamicIntProperty;
 import com.netflix.config.DynamicPropertyFactory;
+import com.netflix.zuul.Attrs;
 import com.netflix.zuul.context.SessionContext;
 import com.netflix.zuul.filters.ZuulFilter;
 import com.netflix.zuul.message.http.HttpHeaderNames;
@@ -31,6 +32,8 @@ import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+import org.w3c.dom.Attr;
 
 /**
  * User: michaels@netflix.com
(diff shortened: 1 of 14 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
```diff
@@ -31,6 +32,8 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+import org.w3c.dom.Attr;
```
artgon, line 36: wrong import?

carl-mastrangelo: fixed

artgon (approved):

artgon:
I like it conceptually but I don't like the name "Attrs". For example, I think `SessionContext` is a lot more descriptive.

carl-mastrangelo:
@artgon  I can make the accessor something like `callAttrs`.  I feel Session is the wrong terminology too (a session persists across multiple calls).   I am using the `Attrs` in other places so I don't think it would be correct to change the type name, just the variable name.

artgon:
What about `RequestAttrs`?

argha-c:
Yeah, given the usage, the name should allude to this being a request level attribute store. Avoid `RequestContext` since that has connotations based on prior usage in other frameworks/libs.

argha-c:
Unrelated to this change but since we're on this class, https://github.com/Netflix/zuul/blob/500295cde056968db4156195d330662c0cc5ae67/zuul-core/src/main/java/com/netflix/zuul/Attrs.java#L65-L67
I think the type conversion should happen on line 67, ie. after the null check, not before.

carl-mastrangelo:
I'm also hesitant to use the word `Request` because its shared for the request and the response, same as SessionContext.

artgon:
So, we need a word that includes both Request and Response but isn't "Session" :)

carl-mastrangelo:
I think `Call` encompasses both.

PR #906, opened 2020-10-08
core: remove some refs to server key

carl-mastrangelo:


carl-mastrangelo:
The idea here is to remove refs to InstanceInfo and Server, so that they can be encapsulated more easily.

PR #910, opened 2020-10-13
core: remove Timings class

carl-mastrangelo:
This timings class is not actually used, and the metrics it exports are not actually recorded anywhere (the values are always zero).   This class poses problems for a separate refactor of the Timing class.


sandy-adi:
@carl-mastrangelo what's the alternative to get timings now?

sandy-adi:
> @carl-mastrangelo what's the alternative to get timings now?

we were actually successfully using this to measure the proxy processing times.

PR #919, opened 2020-11-04
zuul-core: only pick a new server address after deciding to create a …

carl-mastrangelo:
…new connection


artgon:
Only reason I can think of why it's there is if you want to do a DNS lookup on every request.

carl-mastrangelo:
Well, the DNS request would happen, but it wouldn't be used (since the Server object wont change)

PR #923, opened 2020-11-10
rename ALPN based SSL config component

argha-c:
This naming is misleading because we choose the SSL context based on ALPN


artgon (approved):

carl-mastrangelo (approved):
LGTM

PR #929, opened 2020-11-12
add SslContext builder with h2 disabled

argha-c:


zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2Configuration.java
```diff
@@ -77,4 +77,28 @@ public static SslContext configureSSL(SslContextFactory sslContextFactory, Strin
 
         return sslContext;
     }
+
+    /**
+     * This is meant to be use in cases where the server wishes not to advertise h2 as part of ALPN.
+     */
+    private SslContext configureSSLWithH2Disabled(SslContextFactory sslContextFactory, String host) {
+
+        ApplicationProtocolConfig apn = new ApplicationProtocolConfig(
```
artgon, line 86: Why use ALPN if you're not going to use H2?

argha-c: To keep it consistent. I don't see the merit of changing the initialization, based on supported protocols.

PR #930, opened 2020-11-16
[zuul-core] Remove deprecated port attrs. Use ADDR attrs instead.

argha-c:


zuul-core/src/main/java/com/netflix/netty/common/SourceAddressChannelHandler.java
```diff
@@ -89,53 +79,43 @@
...
-     * if present.
+     * The actual local address of the channel, in string form.  If the address is an IPv6 address, the scope identifier
+     * is absent.  Unlike {@link #ATTR_LOCAL_ADDRESS}, this is not overwritten by the Proxy Protocol message if
+     * present.
      *
      * @deprecated Use {@code channel.localAddress()}  instead.
      */
     @Deprecated
-    public static final AttributeKey<String> ATTR_SERVER_LOCAL_ADDRESS = AttributeKey.newInstance("_server_local_address");
+    public static final AttributeKey<String> ATTR_SERVER_LOCAL_ADDRESS = AttributeKey
```
carl-mastrangelo, line 97: Line should be broken at the =

ldayton (commented):
My only concern is that since these were public attributes, there may be other users of zuul who are relying on them. Since each has a clean equivalent replacement, and it's been deprecated for a while, I think it's reasonable to remove.

zuul-core/src/main/java/com/netflix/netty/common/SourceAddressChannelHandler.java
argha-c: Yep, style guide should have picked it. Let me fix.

zuul-core/src/test/java/com/netflix/netty/common/proxyprotocol/ElbProxyProtocolChannelHandlerTest.java
```diff
@@ -163,7 +159,7 @@ public void negotiateProxy_ppv1_ipv4() {
         EmbeddedChannel channel = new EmbeddedChannel();
         channel.pipeline().addLast(ElbProxyProtocolChannelHandler.NAME, new ElbProxyProtocolChannelHandler(registry, true));
         ByteBuf buf = Unpooled.wrappedBuffer(
-                "PROXY TCP4 192.168.0.1 124.123.111.111 10008 443\r\n".getBytes(StandardCharsets.US_ASCII));
+                "PROXY TCP4 192.168.0.1 124.123.111.111 10008 7007\r\n".getBytes(StandardCharsets.US_ASCII));
```
carl-mastrangelo, line 162: This looks messed up.    The destination port is supposed to be the one the client originally connected to, not the final port where it is going.     Thats why the public IP address is matched with the public https port

argha-c: Yeah, you're right. This should be the one that the client connects to, not the listener port. Updated 
The `LOCAL_ADDR` might need renaming, esp. since the meaning changes, when a HAPM is involved.

carl-mastrangelo (approved):
LGTM

PR #932, opened 2020-11-18
[zuul-core] Replace http only channel initializer with ALPN based

argha-c:
This enforces SSL based initialization, with ALPN based swapping of the appropriate protocol handlers.
Edit: This is likely a breaking change, for those relying on the non-SSL based http only initializer today.


artgon:
This is a breaking change and could impact users that don't want to run SSL-based servers.

zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2SslChannelInitializer.java
```diff
@@ -51,18 +51,6 @@
     private final SwallowSomeHttp2ExceptionsHandler swallowSomeHttp2ExceptionsHandler;
     private final String metricId;
 
-
-    /**
-     * Use {@link #Http2SslChannelInitializer(String, ChannelConfig, ChannelConfig, ChannelGroup)} instead.
```
carl-mastrangelo, line 56: The reason this is deprecated (and not deleted) is because this is a backward incompatible change.

argha-c: Well, the reason we deprecate is to eventually be able to remove. 
I see this was only deprecated recently. We should probably make the deprecation window explicit.

carl-mastrangelo: Yeah, I would cut a minor release, then probably make this change.  It's not actually semver, but it's enough to be able to upgrade.

PR #944, opened 2020-11-29
introduce GitHub Actions CI

sullis:
image

image

image


carl-mastrangelo:
Alas, one of the Build teams at Netflix made this change for us (and for the other workflows we have).  Thanks for the PR, though I believe it is now out of date.

For the future, we usually depend on some central teams for Netflix wide OSS support, so we usually have [boring] reasons for holding off on these kinds of updates.

PR #951, opened 2020-12-08
add netty-transport-native-epoll linux-aarch_64

sullis:
add netty-transport-native-epoll with classifier linux-aarch_64
    
This classifier indicates ARM architecture 64 bit
    
Netty 4.1.55 supports the linux-aarch_64 classifier:
https://github.com/netty/netty/pull/10845/files


carl-mastrangelo:
@sullis Can you add that as a runtime dep to your binary?  The reason that linux x64 version is there is mainly to get access to the java classes, not so much to pull in the native deps.

sullis:
> @sullis Can you add that as a runtime dep to your binary? The reason that linux x64 version is there is mainly to get access to the java classes, not so much to pull in the native deps.

Sounds reasonable. I'll close this.

PR #958, opened 2020-12-16
zuul-core: make sure to lock conn counters properly

carl-mastrangelo:
The gauge object returned by Spectator is usually not the same, so it doesn't provide the synchronization needed.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/monitoring/ConnCounter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/monitoring/ConnCounter.java
@@ -109,7 +109,7 @@ public final class ConnCounter {
         Id id = registry.createId(metricBase.name() + '.' + event).withTags(metricBase.tags()).withTags(dimTags);
         Gauge gauge = registry.gauge(id);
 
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/monitoring/ConnCounter.java
```diff
@@ -91,7 +108,8 @@ public void increment(String event) {
         lastCountKey = event;
         Id id = registry.createId(metricBase.name() + '.' + event).withTags(metricBase.tags()).withTags(dimTags);
         Gauge gauge = registry.gauge(id);
-        synchronized (gauge) {
+
+        synchronized (locks[id.hashCode() & LOCK_MASK]) {
```
artgon, line 112: It's going to be a pain to look back at this in the future and try to understand what's going on here.

carl-mastrangelo: I can split it into another method if you would like, but I dont know if I can avoid the lock until the issue (commented above) is fixed.

artgon (approved):

PR #971, opened 2021-01-14
zuul-core: Selectively enable body caching on retries

carl-mastrangelo:
This change modifies the proxy logic to cache request bodies only if there have recently
been throttling events.  This means that if we get back a 503, we more agressively cache
request bodies to be able to retry the request.  Previously, Zuul would cache the body
if the origin was secure, or manually request to cache.  This found to be expensive, since
it results in caching all the time, even if the origin is healthy.

This change makes Zuul look to see if there have been any failures in the past 5 minutes.
If there have, Zuul will cache the body for retry, under the expectation the origin will
recover.  This means the very first request that fails due to throttling cannot be retried,
but subsequent failures can.  After enough time has past from the last throttle, Zuul will
stop caching again.   Zuul can be configured to unconditionally cache by setting this value
to -1.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -164,7 +164,7 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
      * to decide to cache request bodies.
      */
     private static final CachedDynamicLongProperty THROTTLE_MEMORY_SECONDS =
-            new CachedDynamicLongProperty("zuul.proxy.throtle_memory_seconds", Duration.ofMinutes(5).getSeconds());
+            new CachedDynamicLongProperty("zuul.proxy.throttle_memory_seconds", Duration.ofMinutes(5).getSeconds());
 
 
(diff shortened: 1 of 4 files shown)


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -156,6 +159,13 @@
 
     private static final CachedDynamicLongProperty MAX_OUTBOUND_READ_TIMEOUT_MS =
             new CachedDynamicLongProperty("zuul.origin.readtimeout.max", Duration.ofSeconds(90).toMillis());
+    /**
+     * Indicates how long Zuul should remember throttle events for an origin.  As of this writing, throttling is used
+     * to decide to cache request bodies.
+     */
+    private static final CachedDynamicLongProperty THROTTLE_MEMORY_SECONDS =
+            new CachedDynamicLongProperty("zuul.proxy.throtle_memory_seconds", Duration.ofMinutes(5).getSeconds());
```
artgon, line 167: Typo on "throttle". Also, I think 5 minutes is too long of a threshold but can't argue with the idea that we would want to be conservative in estimating this.

carl-mastrangelo: Fixed.   We can easily change this later.

PR #978, opened 2021-01-20
Ensure SessionContext isn't modified when iterating over it

kelnos:
When routing debugging is enabled, Zuul will record changes in the session context after each filter is run.  If the context already has a "routingDebug" property on it, all will be fine.  If not, the comparison code will (as a side effect) add that property.  However, it's added while there is an active iterator over the context.  Once the code tries to fetch the next element from the iterator, a
ConcurrentModificationException is thrown.

This fix is simple, though the ideal fix would probably be to refactor the code to be safer.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/context/Debug.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/context/Debug.java
@@ -121,6 +121,10 @@ public class Debug {
      */
     public static void compareContextState(String filterName, SessionContext context, SessionContext copy) {
         // TODO - only comparing Attributes. Need to compare the messages too.
+
+        // Ensure that the routingDebug property already exists, otherwise we'll have a ConcurrentModificationException below
+        getRoutingDebug(context);
+
         Iterator<String> it = context.keySet().iterator();
         String key = it.next();
         while (key != null) {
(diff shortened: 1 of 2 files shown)


kelnos:
Added a test as well.  Note that if you comment out the fix for the bug, that test fails.

zuul-core/src/main/java/com/netflix/zuul/context/Debug.java
```diff
@@ -121,6 +121,10 @@ public static void addRequestDebug(SessionContext ctx, String line) {
      */
     public static void compareContextState(String filterName, SessionContext context, SessionContext copy) {
         // TODO - only comparing Attributes. Need to compare the messages too.
+
+        // Ensure that the routingDebug property already exists, otherwise we'll have a ConcurrentModificationException below
+        getRoutingDebug(context);
```
sullis, line 126: Perhaps we should add this line too:
```
getRequestDebug(ctx);
```

sullis (changes requested):

zuul-core/src/main/java/com/netflix/zuul/context/Debug.java
kelnos: Not convinced it's necessary, as this code doesn't do anything with the request debug list.  Overall this code should be refactored to avoid mutability and getter side-effects where possible; this is just a simple/quick fix for the issue to stop the bleeding.

sullis: Sounds good to me.

sullis (approved):
lgtm

carl-mastrangelo (approved):
LGTM

carl-mastrangelo:
@kelnos merged, thanks!

PR #992, opened 2021-02-08
[zuul-core] Add counter for conn. close on idle state

argha-c:


artgon:
This seems like a duplicate of this feature: https://github.com/Netflix/zuul/blob/master/zuul-core/src/main/java/com/netflix/netty/common/metrics/ServerChannelMetrics.java#L107

argha-c:
> This seems like a duplicate of this feature: https://github.com/Netflix/zuul/blob/master/zuul-core/src/main/java/com/netflix/netty/common/metrics/ServerChannelMetrics.java#L107

Hmm, indeed. I can port the test, though it'd be nice if channel handlers encapsulated their own metrics imho.

zuul-core/src/test/java/com/netflix/netty/common/CloseOnIdleStateHandlerTest.java
```diff
@@ -0,0 +1,59 @@
...
+    public void incrementCounterOnIdleStateEvent() {
+        final EmbeddedChannel channel = new EmbeddedChannel();
+        channel.pipeline().addLast(new DummyChannelHandler());
+        channel.pipeline().addLast(new CloseOnIdleStateHandler(registry, listener));
+
+        channel.pipeline().context(DummyChannelHandler.class).fireUserEventTriggered(ALL_IDLE_STATE_EVENT);
+
+        verify(counter, times(1)).increment();
+    }
+}H
```
argha-c, line 59: Wut!

zuul-core/src/test/java/com/netflix/netty/common/CloseOnIdleStateHandlerTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CloseOnIdleStateHandlerTest {
+
+    @Mock
+    private Registry registry;
```
carl-mastrangelo, line 37: Consider using a default registry and checking the metric value at the end of the test

argha-c: yep, good call.

argha-c: actually, I don't see an API to lookup the counter for `DefaultRegistry`. 
(I'd have to inject the counter obj itself then, which isn't clean.)

PR #993, opened 2021-02-09
Move conn. throttled counter to its handler

argha-c:


argha-c:
This doesn't need the event passing anymore. Going to clean it up.

PR #1000, opened 2021-02-15
Remove duplicate conns count. Reduce synchronization

argha-c:
The `zuul.conn.client.current.active` effectively employs a similar locking mechanism and it doesn't seem useful to have 2 code paths employing synchronization, for similar goals.
Minor update for this to expose the `nascent` -> `active` counts, which should effectively be a clean replacement for the old metric.


carl-mastrangelo (approved):
LGTM

PR #1001, opened 2021-02-22
Remove unnecessary Governator dependency

artgon:


zuul-guice/src/test/java/com/netflix/zuul/BaseInjectionTest.java
```diff
@@ -0,0 +1,37 @@
...
+import com.netflix.zuul.init.ZuulFiltersModule;
+import org.junit.Before;
+
+/**
+ * Base Injection Test
+ *
+ * @author Arthur Gonigberg
+ * @since February 22, 2021
+ */
+public class BaseInjectionTest {
```
argha-c, line 30: Probably worth conveying this should be required for integration tests only, in the name.

argha-c (approved):

PR #1005, opened 2021-03-01
Modularize ribbon LB deps. Introduce discovery module. 

argha-c:
* `ExecutionContext` is the only dep remaining for removing api dependency on "com.netflix.ribbon:ribbon-loadbalancer"
* `ResolverResult` should ideally model a collection of instances, but exists in the current form, in part since it was simpler to preserve compatibility with current behavior.
* `SimpleMetaInfo` feels extraneous, but also exists to preserve some some existing behavior.


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
```diff
@@ -404,6 +350,23 @@ protected IConnectionPool createConnectionPool(
         );
     }
 
+    final class ServerPoolListener implements ResolverListener<DiscoveryResult> {
+        @Override
+        public void onChange(List<DiscoveryResult> removedSet) {
+                if (!removedSet.isEmpty()) {
```
carl-mastrangelo, line 356: seems a little too indented

zuul-discovery/src/main/java/com/netflix/zuul/resolver/ResolverListener.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+
+/**
+ * @author Argha C
+ * @since 2/25/21
+ *
+ * Listener for resolver updates.
+ */
+public interface ResolverListener<T> {
+
```
carl-mastrangelo, line 28: public interfaces need to have docs for their method.

zuul-discovery/src/main/java/com/netflix/zuul/resolver/Resolver.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @author Argha C
+ * @since 2/25/21
+ *
+ * Resolves a key to a discovery result type.
+ */
+public interface Resolver<T> {
+
+    //TODO(argha-c) Param needs to be typed, once the ribbon LB lookup API is figured out.
+    T resolve(Object key);
```
carl-mastrangelo, line 28: This needs to have a doc describing the key, null-ness, etc.

zuul-discovery/src/main/java/com/netflix/zuul/discovery/SimpleMetaInfo.java
```diff
@@ -0,0 +1,47 @@
...
+import com.netflix.loadbalancer.Server.MetaInfo;
+
+/**
+ * @author Argha C
+ * @since 2/25/21
+ *
+ * placeholder to mimic metainfo for a non-Eureka enabled server.
+ * This exists to preserve compatibility with some current logic, but should be revisited.
+ */
+public class SimpleMetaInfo {
```
carl-mastrangelo, line 28: final?

zuul-discovery/src/main/java/com/netflix/zuul/discovery/NonDiscoveryServer.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.netflix.loadbalancer.Server;
+
+/**
+ * @author Argha C
+ * @since 3/1/21
+ * <p>
+ * This exists merely to wrap a resolver lookup result, that is not discovery enabled.
+ */
+public class NonDiscoveryServer implements ResolverResult {
```
carl-mastrangelo, line 27: final

zuul-discovery/src/main/java/com/netflix/zuul/discovery/NonDiscoveryServer.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final Server server;
+
+    public NonDiscoveryServer(String host, int port) {
+        this.server = new Server(host, port);
```
carl-mastrangelo, line 32: Objects.RequireNonNull? Also it seems like the port should be validated.

zuul-discovery/src/main/java/com/netflix/zuul/discovery/DynamicServerResolver.java
```diff
@@ -0,0 +1,95 @@
...
+        } catch (Exception e) {
+            Throwables.throwIfUnchecked(e);
+            throw new IllegalStateException("Could not instantiate LoadBalancer " + loadBalancerClassName, e);
+        }
+
+        return lb;
+    }
+
+    @VisibleForTesting
+    void onUpdate(List<Server> oldList, List<Server> newList) {
```
carl-mastrangelo, line 87: Some thoughts: 

* It would be surprising if there were multiple values in the list.  I would throw or warn on that.
* Only one of these needs to be put in a set.  You can scan through one list and check if the other [set] has the value.   You can avoid an allocation

zuul-discovery/src/main/java/com/netflix/zuul/discovery/DynamicServerResolver.java
```diff
@@ -0,0 +1,95 @@
...
+        Set<Server> oldSet = new HashSet<>(oldList);
+        Set<Server> newSet = new HashSet<>(newList);
+        final List<DiscoveryResult> discoveryResults = Sets.difference(oldSet, newSet).stream()
```
carl-mastrangelo, line 90: Shouldn't this be the symmetric difference between the two?

zuul-discovery/src/main/java/com/netflix/zuul/discovery/DiscoveryResult.java
```diff
@@ -0,0 +1,183 @@
...
+    public boolean isDiscoveryEnabled() {
+        return server instanceof DiscoveryEnabledServer;
+    }
+
+    @Override
+    public int getPort() {
+        return server.getPort();
+    }
+
+    public String getVIP() {
```
carl-mastrangelo, line 74: Optional: I have been calling this the "target" string, rather than VIP, to get away from NIWs terminology

zuul-discovery/src/main/java/com/netflix/zuul/discovery/DynamicServerResolver.java
argha-c: I would need to verify this since it's existing ribbon behavior. For this refactor, I'd prefer that we didn't change the behavior, at least for now.

zuul-discovery/src/main/java/com/netflix/zuul/discovery/DynamicServerResolver.java
argha-c: The listener interface is designed to only receive the set that was _removed_ from the last update. In effect, that isn't symmetric, since we're only interested in `old - new`.
Arguable if that's the best design since I tried to fit it to the current implementation.

zuul-discovery/src/main/java/com/netflix/zuul/discovery/NonDiscoveryServer.java
argha-c: Done.

zuul-discovery/src/main/java/com/netflix/zuul/discovery/DiscoveryResult.java
argha-c: Done.

zuul-core/build.gradle
```diff
@@ -14,12 +14,14 @@ dependencies {
     api "com.netflix.spectator:spectator-api:0.123.1"
     api "com.netflix.netflix-commons:netflix-commons-util:0.3.0"
 
+    api project(":zuul-discovery")
```
artgon, line 17: If you want the dependency restriction to work (i.e. not be transitive), this should be `implementation` rather than `api`.

argha-c: the ribbon deps in `zuul-discovery` are `implementation`. That effectively means anything pulling in `zuul-discovery` won't have access to those types, which is what we want.

We can probably make this itself restrictive, but that'll apply to the new types defined in discovery, not in ribbon.

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
```diff
@@ -404,6 +350,24 @@ protected IConnectionPool createConnectionPool(
         );
     }
 
+    final class ServerPoolListener implements ResolverListener<DiscoveryResult> {
+
+        @Override
+        public void onChange(List<DiscoveryResult> removedSet) {
+            if (!removedSet.isEmpty()) {
+                LOG.debug("Removing connection pools for missing servers. name = " + originName
+                        + ". " + removedSet.size() + " servers gone.");
```
carl-mastrangelo, line 359: use `{}` and an extra arg?

zuul-core/src/main/java/com/netflix/zuul/niws/RequestAttempt.java
```diff
@@ -24,9 +24,9 @@
 import com.netflix.appinfo.InstanceInfo;
```
carl-mastrangelo, line 24: I may be mistaken, but I thin this comes from discovery too.

argha-c: this comes from a different package, ie, `eureka-client` and not one of the ribbon deps. But yes, we want to move away from it, at some point, ideally.

zuul-discovery/src/main/java/com/netflix/zuul/resolver/Resolver.java
argha-c: Done.

zuul-discovery/src/main/java/com/netflix/zuul/resolver/ResolverListener.java
argha-c: Done.

PR #1006, opened 2021-03-02
Upgrade nebula.netflixoss to replace bintray publication and update TravisCI secrets

rpalcolea:
Hi folks,

This upgrades to Nebula NetflixOSS which will publish to Nebula's OSS repos and Maven Central.

Repositories based on status:

* https://netflixoss.jfrog.io/artifactory/maven-oss-snapshots
* https://netflixoss.jfrog.io/artifactory/maven-oss-candidates
* https://netflixoss.jfrog.io/artifactory/maven-oss-releases

This also takes care of new TravisCI secrets required to accomplish publishing tasks


build.gradle
```diff
@@ -1,6 +1,6 @@
 
 plugins {
-    id 'nebula.netflixoss' version '8.8.1'
+    id 'nebula.netflixoss' version '9.1.0'
```
carl-mastrangelo, line 3: You'll need to run the dependency lock command in this PR for this to take effect.

buildViaTravis.sh
```diff
@@ -6,15 +6,15 @@ if [ "$TRAVIS_PULL_REQUEST" != "false" ] || [ "$GRADLE_PUBLISH" == "false" ]; th
   ./gradlew build
 elif [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_TAG" == "" ]; then
   echo -e 'Build Branch with Snapshot => Branch ['$TRAVIS_BRANCH']'
-  ./gradlew -Prelease.travisci=true -PbintrayUser="${bintrayUser}" -PbintrayKey="${bintrayKey}" -PsonatypeUsername="${sonatypeUsername}" -PsonatypePassword="${sonatypePassword}" -Prelease.scope=patch build snapshot
+  ./gradlew -Prelease.travisci=true -PnetflixOss.username=$NETFLIX_OSS_REPO_USERNAME -PnetflixOss.password=$NETFLIX_OSS_REPO_PASSWORD  -Psonatype.signingPassword=$NETFLIX_OSS_SIGNING_PASSWORD -Prelease.scope=patch build snapshot
```
carl-mastrangelo, line 9: These args need to be wrapped in quotes.

build.gradle
rpalcolea: We don't lock buildscript classpath dependencies with nebula lock so this isn't necessary. Running `generateLock saveLock` only upgrades jmh and mockito-core.

Will add them

buildViaTravis.sh
rpalcolea: not necessarily, take https://github.com/Netflix/netflixoss-sample-project-travisci/blob/main/gradle/buildViaTravis.sh#L21 as example

Will add it for consistency with previous versions

rpalcolea:
Locks are updated and added the missing double quotes

buildViaTravis.sh
carl-mastrangelo: Thanks.   Ideally these scripts would pass shellcheck, but it there are many gaps in our consistency.  As long as the problem doesn't get worse, I'm okay.

carl-mastrangelo (approved):
LGTM

PR #1030, opened 2021-03-18
Replace JCenter with MavenCentral

rpalcolea:
Hi folks,

This is to stop using `jcenter()` as it is going away. `mavenCentral()` is preferred


carl-mastrangelo (approved):
LGTM

PR #1031, opened 2021-03-23
Simplify read timeout functionality and move to using Duration

artgon:
This is a fairly large refactoring of how timeouts work in `ProxyEndpoint`. I've simplified the logic by removing the ability to use `ExecutionListeners` (from Ribbon) to modify timeout settings. By removing that feature, we no longer need to set the timeout on the `ClientConfig` object repeatedly. Now, we can have a single `Duration` field that we re-compute on each attempt, with no machinery required to pull things in and out of the request config and origin config. In fact, those configs remain completely untouched through proxying, meaning we also don't have to reset them when we're done with each computation.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -379,6 +379,7 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
             attemptToChosenHostMap = new HashMap<>();
         }
 
+        // the chosen server can be null in the case of a timeout exception that skips acquiring a new origin connection
         if (chosenServer.get() != null) {
             String ipAddr = origin.getIpAddrFromServer(chosenServer.get());
             if (ipAddr != null) {
@@ -405,7 +406,11 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
         try {
             attemptNum += 1;
 
-            // compute how much time we have left for this attempt
+            /*
+             * Before connecting to the origin, we need to compute how much time we have left for this attempt. This
+             * method is also intended to validate deadline and timeouts boundaries for the request as a whole and could
+             * throw an exception, skipping the logic below.
+             */
             timeLeftForAttempt = originTimeoutManager.computeReadTimeout(zuulRequest, attemptNum);
 
             currentRequestStat = createRequestStat();
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -26,12 +26,12 @@ import static com.netflix.zuul.stats.status.ZuulStatusCategory.FAILURE_ORIGIN_TH
 import static com.netflix.zuul.stats.status.ZuulStatusCategory.SUCCESS;
 import static com.netflix.zuul.stats.status.ZuulStatusCategory.SUCCESS_LOCAL_NO_ROUTE;
 import static com.netflix.zuul.stats.status.ZuulStatusCategory.SUCCESS_NOT_FOUND;
+
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Sets;
 import com.google.errorprone.annotations.ForOverride;
 import com.netflix.client.ClientException;
-import com.netflix.client.config.IClientConfig;
 import com.netflix.client.config.IClientConfigKey;
 import com.netflix.client.config.IClientConfigKey.Keys;
 import com.netflix.config.CachedDynamicLongProperty;
@@ -139,8 +139,7 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
     protected MethodBinding<?> methodBinding;
     protected HttpResponseMessage zuulResponse;
     protected boolean startedSendingResponseToClient;
-    // this is an Object because it can be a String or Integer depending on timeout config
-    protected Object originalReadTimeout;
+    protected Duration timeLeftForAttempt;
 
     /* Individual retry related state */
     private volatile PooledConnection originConn;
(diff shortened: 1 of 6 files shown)


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -389,11 +378,14 @@ private void storeAndLogOriginRequestInfo() {
         if (attemptToChosenHostMap == null) {
             attemptToChosenHostMap = new HashMap<>();
         }
-        String ipAddr = origin.getIpAddrFromServer(chosenServer.get());
-        if (ipAddr != null) {
-            attemptToIpAddressMap.put(attemptNum, ipAddr);
-            eventProps.put(CommonContextKeys.ZUUL_ORIGIN_ATTEMPT_IPADDR_MAP_KEY, attemptToIpAddressMap);
-            context.put(CommonContextKeys.ZUUL_ORIGIN_ATTEMPT_IPADDR_MAP_KEY, attemptToIpAddressMap);
+
+        if (chosenServer.get() != null) {
```
argha-c, line 382: You probably want to check for the sentinel value of `DiscoveryResult.EMPTY`. This is unlikely to be null, as part of resolver lookup.

zuul-core/src/main/java/com/netflix/zuul/netty/timeouts/OriginTimeoutManager.java
```diff
@@ -124,4 +99,14 @@ private Long parseReadTimeoutMs(Object p) {
             return null;
         }
     }
+
+    @Nullable
+    private Object getRequestReadTimeout(HttpRequestMessage zuulRequest, Long defaultTimeout) {
+        return getRequestClientConfig(zuulRequest).getProperty(ReadTimeout, defaultTimeout);
```
carl-mastrangelo, line 105: I know it's not your fault, but `ReadTimeout`, that is not good.

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
carl-mastrangelo, line 382: When would this be null, and why didnt the previous code check for it?

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ClientTimeoutHandler.java
```diff
@@ -55,10 +56,10 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
         @Override
         public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
             try {
-                final Integer timeout = ctx.channel().attr(ORIGIN_RESPONSE_READ_TIMEOUT).get();
+                final Duration timeout = ctx.channel().attr(ORIGIN_RESPONSE_READ_TIMEOUT).get();
                 if (timeout != null && msg instanceof LastHttpContent) {
                     promise.addListener(e -> {
-                        LOG.debug("[{}] Adding read timeout handler: {}", ctx.channel().id(), timeout);
+                        LOG.debug("[{}] Adding read timeout handler: {}", ctx.channel().id(), timeout.toMillis());
```
carl-mastrangelo, line 62: personal preference, but I find the non-`toMillis` more helpful to read.

zuul-core/src/main/java/com/netflix/zuul/netty/timeouts/OriginTimeoutManager.java
```diff
@@ -78,25 +58,20 @@ public IClientConfig getRequestClientConfig(HttpRequestMessage zuulRequest) {
...
      */
-    protected Duration getReadTimeout(IClientConfig requestConfig, int attemptNum) {
+    public Duration computeReadTimeout(HttpRequestMessage request, int attemptNum) {
         Long noTimeout = null;
         // TODO(carl-mastrangelo): getProperty is deprecated, and suggests using the typed overload `get`.   However,
         //  the value is parsed using parseReadTimeoutMs, which supports String, implying not all timeouts are Integer.
         //  Figure out where a string ReadTimeout is coming from and replace it.
         Long originTimeout = parseReadTimeoutMs(getOriginReadTimeout(noTimeout));
-        Long requestTimeout = parseReadTimeoutMs(getRequestReadTimeout(requestConfig, noTimeout));
+        Long requestTimeout = parseReadTimeoutMs(getRequestReadTimeout(request, noTimeout));
```
carl-mastrangelo, line 74: I feel this is a code smell.  The `request` is being passed into `getRequestReadTimeout`, but only the config is being used.   Rather than `getRequestReadTimeout` being responsible for extracting the config, it seems like it should be passed in, which would mean extracting it just above this line, and passing the config in.

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ClientTimeoutHandler.java
artgon: Was thinking it would be simpler since most configs are in millis.

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
artgon: This is purely a null check to prevent NPE.

artgon: This can now happen because `computeReadTimeout` is intended to also validate conditions. For example, it could throw a timeout exception and thus no server will be chosen.

zuul-core/src/main/java/com/netflix/zuul/netty/timeouts/OriginTimeoutManager.java
artgon: Can you elaborate?

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
carl-mastrangelo: Consider adding a comment, it's not obvious from reading the code that case can happen.

zuul-core/src/main/java/com/netflix/zuul/netty/timeouts/OriginTimeoutManager.java
carl-mastrangelo: The camelCasing is off for this.

artgon: It's a statically imported constant, I can use the full class if you think that's clearer.

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
argha-c: Okay. Probably worth adding a comment for when that's likely to happen.

zuul-core/src/test/java/com/netflix/zuul/netty/timeouts/OriginTimeoutManagerTest.java
```diff
@@ -0,0 +1,121 @@
...
+        originConfig.set(CommonClientConfigKey.ReadTimeout, 100);
+
+        Duration timeout = originTimeoutManager.computeReadTimeout(request, 1);
+
+        assertEquals(100, timeout.toMillis());
+    }
+
+    @Test
+    public void computeReadTimeout_bolth_requestLower() {
+        requestConfig.set(CommonClientConfigKey.ReadTimeout, 1000);
```
argha-c, line 118: Shouldn't this be lower per the test name?

artgon: ah yeah, copy pasta

argha-c (approved):
LGTM

PR #1046, opened 2021-04-14
downgrade ClientResponseWriter logging

sullis:
This PR changes LOG.warn to LOG.info in ClientResponseWriter

Motivation:

We have observed a large number of WARN messages in production.
The log message that occurs most frequently is

```
Received complete event while still handling the request. With reason: CLOSE
```

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientResponseWriter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientResponseWriter.java
@@ -247,7 +247,7 @@ public class ClientResponseWriter extends ChannelInboundHandlerAdapter {
             }
             else {
                 if (isHandlingRequest) {
-                    LOG.warn("Received complete event while still handling the request. With reason: " + reason.name() + " -- " +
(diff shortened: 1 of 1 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientResponseWriter.java
```diff
@@ -247,7 +247,7 @@ else if (evt instanceof CompleteEvent) {
             }
             else {
                 if (isHandlingRequest) {
-                    LOG.warn("Received complete event while still handling the request. With reason: " + reason.name() + " -- " +
+                    LOG.info("Received complete event while still handling the request. With reason: " + reason.name() + " -- " +
```
argha-c, line 250: I think this is likely due to clients that employ http pipelining, which zuul doesn't support. #1045 should ensure that the `reason` in the message is populated accurately, when that is indeed the case.

That said, the downgrade LGTM.

sullis: > That said, the downgrade LGTM. Probably consider DEBUG.

I just updated the code. It now uses LOG.debug

argha-c:
LGTM

PR #1050, opened 2021-04-20
zuul-guice: update to Guice 5

carl-mastrangelo:
This is needed to run on later JDKs than 11, as it uses a better maintained classfile generator.   From my reading, it is API compatible with Guice 4.2.

"

### DIFF
--- a/zuul-guice/build.gradle
+++ b/zuul-guice/build.gradle
@@ -2,7 +2,7 @@ apply plugin: "java-library"
 
 dependencies {
     implementation project(":zuul-core")
-    api(group: 'com.google.inject', name: 'guice', version: "4.2.3", classifier: "no_aop")
+    api(group: 'com.google.inject', name: 'guice', version: "5.0.1")
     implementation 'commons-configuration:commons-configuration:1.8'
 
     testImplementation libraries.junit,
--- a/zuul-core/dependencies.lock
+++ b/zuul-core/dependencies.lock
@@ -271,7 +271,7 @@
             "locked": "1.68"
         },
         "org.mockito:mockito-core": {
-            "locked": "3.8.0"
+            "locked": "3.9.0"
         },
         "org.openjdk.jmh:jmh-core": {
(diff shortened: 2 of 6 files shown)


zuul-guice/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: "java-library"
 
 dependencies {
     implementation project(":zuul-core")
-    api(group: 'com.google.inject', name: 'guice', version: "4.2.3", classifier: "no_aop")
+    api(group: 'com.google.inject', name: 'guice', version: "5.0.1")
```
argha-c, line 5: From the release notes, looks like they introduced `-Dguice_bytecode_gen_option=DISABLED` to stick to not using AOP based code. Thoughts?

carl-mastrangelo: the aop thing was to work around JDK11 problems iirc.  Using the supported byte code reader/generator should obviate the need for aop specific stuff.

argha-c: Right, so probably worth making it explicit setting that flag internally, imo.

carl-mastrangelo: Well, what I am saying is that the sole reason for the classifier was to work around a bug with how cglib read and generated classes.  Now that cglib is not being used, there is no reason to be for or against codegen.  Those generated classes tend to have performance optimizations, so I'd prefer to have them, as long as they dont cause trouble.    In other words, I don't think we need to set that System property.

carl-mastrangelo:
We can continue to discuss the sysprop.  I am going to submit, but please continue to express any latent concerns you have.

zuul-guice/build.gradle
argha-c: cglib removal only changes what gets used for bytecode generation. My point being, as of today, `no_aop` implies that we don't intend on supporting aop based interception. This changes that behavior.

I'm not a fan of bytecode generation being a requirement for DI.

PR #1060, opened 2021-05-12
Block requests with multiple Host headers

jfernandez:


zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -159,6 +158,19 @@ private void channelReadInternal(final ChannelHandlerContext ctx, Object msg) th
                         ZuulStatusCategory.FAILURE_CLIENT_BAD_REQUEST);
                 zuulRequest.getContext().setError(ze);
                 zuulRequest.getContext().setShouldSendErrorResponse(true);
+            } else if (zuulRequest.getHeaders().getAll(HttpHeaderNames.HOST.toString()).size() > 1) {
+                LOG.debug(
+                        "Duplicate Host headers. clientRequest = {} , uri = {}, info = {}",
+                        clientRequest.toString(),
```
carl-mastrangelo, line 164: s/.toString()//

carl-mastrangelo (approved):
LGTM

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -159,6 +158,19 @@ private void channelReadInternal(final ChannelHandlerContext ctx, Object msg) th
...
+                        clientRequest,
+                        clientRequest.uri(),
+                        ChannelUtils.channelInfoForLogging(ctx.channel()));
+                final ZuulException ze = new ZuulException("Duplicate Host headers");
+                ze.setStatusCode(HttpResponseStatus.BAD_REQUEST.code());
+                StatusCategoryUtils.setStatusCategory(
+                        zuulRequest.getContext(),
+                        ZuulStatusCategory.FAILURE_CLIENT_BAD_REQUEST);
+                zuulRequest.getContext().setError(ze);
+                zuulRequest.getContext().setShouldSendErrorResponse(true);
```
argha-c, line 173: This will continue the filter chain processing, and actually write a response to the client.
Is that desirable?

jfernandez: Yes, I believe so, but interested in hearing your thoughts. The other option would be to close the connection without a response (we do this a few lines above). It seems best to let the client know what they did wrong.

argha-c: I'm not aware of a good spec for this case. 
My general preference is to not continue filter processing and expend resources writing back a response, when we treat a request as malicious.

jfernandez: For this change, I'm inclined to provide the client a 400 response given that it could impact non-malicious requests. For example, a client or upstream proxy may be accidentally adding the Host header twice. Closing the connection without a response would make it very difficult to troubleshoot this change when it rolls out. I think the better client experience outweighs the extra resource spend here.

argha-c (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -159,6 +158,19 @@ private void channelReadInternal(final ChannelHandlerContext ctx, Object msg) th
                         ZuulStatusCategory.FAILURE_CLIENT_BAD_REQUEST);
                 zuulRequest.getContext().setError(ze);
                 zuulRequest.getContext().setShouldSendErrorResponse(true);
+            } else if (zuulRequest.getHeaders().getAll(HttpHeaderNames.HOST.toString()).size() > 1) {
```
argha-c, line 161: This should be `HOST.getName()`

jfernandez: HOST is an `AsciiString`, that method doesn't resolve for it.

argha-c (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
argha-c: edit: `com.netflix.zuul.message.http.HttpHeaderNames`

PR #1062, opened 2021-05-15
Updating log4j properties file in zuul-sample

kashish9034z:
**What is the Issue ?**

Existing log4j properties file is not being picked by log4j and there are no logs produced to console.

**Why does this issue exist ?**

Log4j2 uses a different format of properties file which is explained on apache's [website](https://logging.apache.org/log4j/log4j-2.4/manual/configuration.html#Loggers)

image

**ChangeLog-**

- Adding Log4j2.properties in the new format
- Removing existing log4j properties file

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/util/HttpRequestBuilder.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/util/HttpRequestBuilder.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ *      Licensed under the Apache License, Version 2.0 (the "License");
+ *      you may not use this file except in compliance with the License.
+ *      You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *      Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


zuul-sample/src/main/resources/log4j2.properties
```diff
@@ -0,0 +1,23 @@
+status = error
+name = Log4j2PropertiesConfig
+
+appenders = console
+
+appender.console.type = Console
+appender.console.name = LogToConsole
```
argha-c, line 7: Typically, this naming should be the stream being logged, ex: `stderr`.

zuul-sample/src/main/resources/log4j2.properties
```diff
@@ -0,0 +1,23 @@
...
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %d %-5p %c [%t] %m%n
+
+rootLogger.level = info
+rootLogger.appenderRefs = stdout
```
argha-c, line 12: Is this appender defined?

argha-c (changes requested):
Thanks for updating this. LGTM but some comments.

zuul-sample/src/main/resources/log4j2.properties
kashish9034z: Updated appender name

PR #1066, opened 2021-05-20
zuul-core: Add overload for OriginName trust

carl-mastrangelo:


zuul-core/src/main/java/com/netflix/zuul/origins/OriginName.java
```diff
@@ -44,30 +42,36 @@
...
     public static OriginName fromVip(String vip, String niwsClientName) {
-        return new OriginName(niwsClientName, vip, VipUtils.extractUntrustedAppNameFromVIP(vip), false);
+        return fromVipAndApp(vip, VipUtils.extractUntrustedAppNameFromVIP(vip), niwsClientName);
     }
 
-    @CheckReturnValue
-    public OriginName withTrustedAuthority(String authority) {
-        return new OriginName(niwsClientName, target, authority, true);
+    public static OriginName fromVipAndApp(String vip, String appName) {
+        return fromVipAndApp(vip, appName, vip);
```
argha-c, line 63: This isn't obvious without a comment explaining why the vip and niws name may or may not be the same.

zuul-core/src/test/java/com/netflix/zuul/origins/OriginNameTest.java
```diff
@@ -27,13 +26,9 @@
 @RunWith(JUnit4.class)
 public class OriginNameTest {
```
argha-c, line 27: This needs more unit tests, at least:
* verification of equality check
* Tests for when niws name and vip are different
*

zuul-core/src/main/java/com/netflix/zuul/origins/OriginName.java
carl-mastrangelo: Followed up here: https://github.com/Netflix/zuul/pull/1067

zuul-core/src/test/java/com/netflix/zuul/origins/OriginNameTest.java
carl-mastrangelo: See above PR.

PR #1069, opened 2021-05-21
chore(build): add JDK 16 to build matrix

sullis:


carl-mastrangelo:
This depends on updating Nebula to be on Gradle 6.9, or on 7.0, both of which are not yet available I think.  cc @rpalcolea

PR #1084, opened 2021-07-29
Register the allocator memory metric polled gauge only once

jfernandez:
We previously registered a polled gauge for each port. If you register a polled gauge multiple times with the same id, then it will sum the corresponding values. This was causing our memory metrics to report x N as the actual value (where N is number of ports).

All channels should be sharing the same allocator instance, so we only need to register the gauge once. Querying this metric also acquires a lock on each arena.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
@@ -203,6 +203,19 @@ public class Server
                 addressesToChannels.put(requestedNamedAddr.withNewSocket(chan.localAddress()), chan);
                 allBindFutures.add(nettyServerFuture);
             }
+
+            // All channels should share a single ByteBufAllocator instance.
+            // Add metrics to monitor that allocator's memory usage.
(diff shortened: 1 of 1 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -203,6 +203,19 @@ public void start()
                 addressesToChannels.put(requestedNamedAddr.withNewSocket(chan.localAddress()), chan);
                 allBindFutures.add(nettyServerFuture);
             }
+
+            // All channels should share a single ByteBufAllocator instance.
+            // Add metrics to monitor that allocator's memory usage.
+            if (!allBindFutures.isEmpty()) {
```
carl-mastrangelo, line 209: Should this conditional be inverted?

jfernandez: We want to check here that the `allbindFutures` list is **not** empty because in line 201 we fetch the first entry. The conditional looks correct to me.

carl-mastrangelo (approved):
LGTM

PR #1086, opened 2021-08-03
Add Typesafe session context accessors 

carl-mastrangelo:
These accessors are more typesafe than the string based ones, and make cross referencing usage easier.  In addition they are strict about null safety, making it hard for unexpected nulls to be present in the session context.


artgon:
Nice! 👏

PR #1094, opened 2021-08-12
Fix ConcurrentModificationException when calling getBodyLength

jfernandez:
Calling ZuulMessageImpl.getBodyLength can sometimes trigger this exception. My assumption is that it happens when the `bodyChunks` list is modified while the method is iterating through it.

```
java.util.ConcurrentModificationException: 
  at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1013)
  at java.util.ArrayList$Itr.next(ArrayList.java:967)
  at com.netflix.zuul.message.ZuulMessageImpl.getBodyLength(ZuulMessageImpl.java:170)
  at com.netflix.zuul.message.http.HttpRequestMessageImpl.getBodyLength(HttpRequestMessageImpl.java:218)
  ```

I didn't use `getBodyContents` because that wraps it in `Collections.unmodifiableList` which doesn't prevent ConcurrentModificationExceptions if the underlying list is modified.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
@@ -17,6 +17,7 @@ package com.netflix.zuul.message;
 
 import com.google.common.base.Charsets;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
 import com.netflix.config.DynamicIntProperty;
 import com.netflix.config.DynamicPropertyFactory;
 import com.netflix.zuul.context.SessionContext;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java [hunk 1]
```diff
@@ -167,7 +168,7 @@ public String getBodyAsText() {
     @Override
     public int getBodyLength() {
         int size = 0;
-        for (final HttpContent chunk : bodyChunks) {
+        for (final HttpContent chunk : ImmutableList.copyOf(bodyChunks)) {
```
carl-mastrangelo, line 171: Immutable list is just going to iterate through it as well, I dont think this would fix it.

PR #1095, opened 2021-08-17
Refactor PassportLoggingHandler to reduce unnecessary heap allocation

jfernandez:
The "slow response/request processing" INFO level adds too much noise when a slow GC pause slows down our filter processing.

This PR:
* Downgrades slow response/request processing logging to DEBUG
* Refactors PassportLoggingHandler to leverage isDebugEnabled and remove unnecessary heap allocation and timing calculation.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
@@ -99,21 +99,31 @@ public class PassportLoggingHandler extends ChannelInboundHandlerAdapter
                     + ", toplevelid = " + topLevelRequestId
                     + ", req = " + request.getInfoForLogging()
                     + ", passport = " + String.valueOf(passport));
-        }
 
-        // Some logging of session states if certain criteria match:
-        if (LOG.isInfoEnabled()) {
-            if (passport.wasProxyAttempt()) {
-
-                if (passport.findStateBackwards(PassportState.OUT_RESP_LAST_CONTENT_SENDING) == null) {
-                    incompleteProxySessionCounter.increment();
-                    LOG.info("Incorrect final state! toplevelid = " + topLevelRequestId + ", " + ChannelUtils.channelInfoForLogging(channel));
-                }
+            StartAndEnd inReqToOutResp = passport.findFirstStartAndLastEndStates(PassportState.IN_REQ_HEADERS_RECEIVED, PassportState.OUT_REQ_LAST_CONTENT_SENT);
+            if (passport.calculateTimeBetween(inReqToOutResp) > WARN_REQ_PROCESSING_TIME_NS.get()) {
+                LOG.debug("Request processing took longer than threshold! toplevelid = " + topLevelRequestId + ", "
+                        + ChannelUtils.channelInfoForLogging(channel));
             }
 
-            if (! passport.wasProxyAttempt()) {
-                if (ctx != null && !isHealthcheckRequest(request)) {
-                    // Why did we fail to attempt to proxy this request?
+            StartAndEnd inRespToOutResp = passport.findLastStartAndFirstEndStates(PassportState.IN_RESP_HEADERS_RECEIVED, PassportState.OUT_RESP_LAST_CONTENT_SENT);
+            if (passport.calculateTimeBetween(inRespToOutResp) > WARN_RESP_PROCESSING_TIME_NS.get()) {
+                LOG.debug("Response processing took longer than threshold! toplevelid = " + topLevelRequestId + ", "
+                        + ChannelUtils.channelInfoForLogging(channel));
+            }
+        }
+
+        if (passport.wasProxyAttempt()) {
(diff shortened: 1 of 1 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
```diff
@@ -99,21 +99,31 @@ private void logPassport(Channel channel)
                     + ", toplevelid = " + topLevelRequestId
                     + ", req = " + request.getInfoForLogging()
                     + ", passport = " + String.valueOf(passport));
-        }
```
jfernandez, line 102: There is a `f (LOG.isDebugEnabled()) {` at the top of this block

zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
```diff
@@ -99,21 +99,31 @@ private void logPassport(Channel channel)
...
 
-        // Some logging of session states if certain criteria match:
-        if (LOG.isInfoEnabled()) {
-            if (passport.wasProxyAttempt()) {
+            StartAndEnd inReqToOutResp = passport.findFirstStartAndLastEndStates(PassportState.IN_REQ_HEADERS_RECEIVED, PassportState.OUT_REQ_LAST_CONTENT_SENT);
```
jfernandez, line 103: There is a `if (LOG.isDebugEnabled()) {` conditional at the top of this block.

zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
```diff
@@ -99,21 +99,31 @@ private void logPassport(Channel channel)
...
+            if (passport.findStateBackwards(PassportState.OUT_RESP_LAST_CONTENT_SENDING) == null) {
+                incompleteProxySessionCounter.increment();
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Incorrect final state! toplevelid = " + topLevelRequestId + ", " + ChannelUtils.channelInfoForLogging(channel));
+                }
+            }
+        } else {
+            if (ctx != null && !isHealthcheckRequest(request)) {
+                // Why did we fail to attempt to proxy this request?
+                if (LOG.isDebugEnabled()) {
```
argha-c, line 126: This is redundant, and no longer necessary with the log4j implementation of slf4j. 
It'd actually make sense to clean up the conditionals here.

jfernandez: The input to LOG.debug below is evaluated before the logger method itself does the log level check. The advice I received is to wrap the call with this conditional if we're constructing a log message dynamically.

zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
```diff
@@ -99,21 +99,31 @@ private void logPassport(Channel channel)
...
                     + ", req = " + request.getInfoForLogging()
                     + ", passport = " + String.valueOf(passport));
-        }
 
-        // Some logging of session states if certain criteria match:
-        if (LOG.isInfoEnabled()) {
-            if (passport.wasProxyAttempt()) {
+            StartAndEnd inReqToOutResp = passport.findFirstStartAndLastEndStates(PassportState.IN_REQ_HEADERS_RECEIVED, PassportState.OUT_REQ_LAST_CONTENT_SENT);
+            if (passport.calculateTimeBetween(inReqToOutResp) > WARN_REQ_PROCESSING_TIME_NS.get()) {
+                LOG.debug("Request processing took longer than threshold! toplevelid = " + topLevelRequestId + ", "
```
argha-c, line 105: If we actually want to improve allocation, avoiding string concatenation is a much better low hanging fruit here imo.

jfernandez: We want to primarily reduce logging noise, and secondarily improve allocation.

zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
argha-c: Hmm, I left a separate comment around string concat but there have been optimizations there. 
It's difficult to discuss this without an allocation profile imo.

zuul-core/src/main/java/com/netflix/zuul/netty/insights/PassportLoggingHandler.java
carl-mastrangelo: The PR description calls out that this is for the purpose of reducing allocation.

mc639: what would be a better alternative to not doing String concatenation, using a ```StringBuilder``` or ```StringBuffer``` to build the string only once and then ```LOG.debug``` once?

PR #1096, opened 2021-08-17
Use a metric instead of logging excessive filter execution time

jfernandez:
The information logged here can be useful, but when Zuul slows down these messages will drown out more useful log entries (which may point at the root cause). This data is more useful as a metric.


carl-mastrangelo (approved):
LGTM

carl-mastrangelo (approved):

PR #1098, opened 2021-08-23
Optimized Logic to strip Malicious Header Characters like '\r' & '\n'.

mc639:
### Notes 

```Optimized Logic to strip Malicious Header Characters like '\r' & '\n'.```

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/util/HttpUtils.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/util/HttpUtils.java
@@ -69,7 +69,7 @@ public class HttpUtils
             return null;
         }
         xForwardedFor = xForwardedFor.trim();
-        String tokenized[] = xForwardedFor.split(",");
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: zuul-core/src/main/java/com/netflix/zuul/util/HttpUtils.java [hunk 1]
```diff
@@ -104,16 +104,21 @@ public static boolean acceptsGzip(Headers headers) {
...
+
         if (input == null) {
             return null;
         }
-        // TODO(carl-mastrangelo): implement this more efficiently.
-        for (char c : MALICIOUS_HEADER_CHARS) {
-            if (input.indexOf(c) != -1) {
-                input = input.replace(Character.toString(c), "");
+
+        for(char inputChar : input.toCharArray()){
```
carl-mastrangelo, line 113: This would result in the input string being rebuilt each call to `stripMaliciousHeaderChars`.  The previous implementation had a fast path for no invalid characters.

PR #1168, opened 2022-01-05
Bypass deflate content from gzip

artgon:
Fixes an issue where `deflate` encoded bodies would be gzipped, creating a double encoding.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/common/GZipResponseFilter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/common/GZipResponseFilter.java
@@ -78,7 +78,7 @@ public class GZipResponseFilter extends HttpOutboundSyncFilter
 
         // Check the headers to see if response is already gzipped.
         final Headers respHeaders = response.getHeaders();
-        boolean isResponseGzipped = HttpUtils.isGzipped(respHeaders);
+        boolean isResponseGzipped = HttpUtils.isGzipped(respHeaders) || HttpUtils.isDeflated(respHeaders);
 
         // Decide what to do.;
         final boolean shouldGzip = isGzippableContentType(response) && isGzipRequested && !isResponseGzipped && isRightSizeForGzip(response);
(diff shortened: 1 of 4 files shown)


zuul-core/src/main/java/com/netflix/zuul/filters/common/GZipResponseFilter.java
```diff
@@ -78,7 +78,7 @@ public boolean shouldFilter(HttpResponseMessage response) {
 
         // Check the headers to see if response is already gzipped.
         final Headers respHeaders = response.getHeaders();
-        boolean isResponseGzipped = HttpUtils.isGzipped(respHeaders);
+        boolean isResponseGzipped = HttpUtils.isGzipped(respHeaders) || HttpUtils.isDeflated(respHeaders);
```
kyagna, line 81: Minor nitpick: rename the variable to isResponseCompressed since we are checking for 2 different compression types.

zuul-core/src/test/java/com/netflix/zuul/filters/common/GZipResponseFilterTest.java
```diff
@@ -102,6 +104,65 @@ public void prepareResponseBody_NeedsGZipping() throws Exception {
...
+        // Check Content-Length header has been removed
+        assertEquals(0, result.getHeaders().getAll("Content-Length").size());
+    }
+
+    @Test
+    public void prepareResponseBody_alreadyZipped() throws Exception {
+        originalRequestHeaders.set("Accept-Encoding", "gzip,deflate");
+
+        byte[] originBody = "blah".getBytes();
+        response.getHeaders().set("Content-Length", Integer.toString(originBody.length));
```
argha-c, line 149: For correctness, we should also set `Content-Type` on the response.

zuul-core/src/main/java/com/netflix/zuul/util/HttpUtils.java
```diff
@@ -84,14 +85,23 @@ public static String extractClientIpFromXForwardedFor(String xForwardedFor) {
      * @return true if the content-encoding param containg gzip
      */
     public static boolean isGzipped(String contentEncoding) {
-        return contentEncoding.contains("gzip");
+        return contentEncoding.contains(HttpHeaderValues.GZIP.toString());
+    }
+
+    public static boolean isDeflated(String contentEncoding) {
```
argha-c, line 91: nit: Unrelated to your change, these should probably be `private`. The public interface should accept the actual headers and not the raw String(like in `isGzipped`)

argha-c (approved):

PR #1180, opened 2022-01-26
This fixes an issue where Zuul does not correctly handle absolute URIs from Netty

artgon:
As per https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.2 the path segment of an inbound request can have an absolute URI in it. This fixes an issue where Zuul assumes that it's always a relative URI.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
@@ -89,6 +89,7 @@ public class ClientRequestReceiver extends ChannelDuplexHandler {
     private static final String SCHEME_HTTPS = "https";
 
     // via @stephenhay https://mathiasbynens.be/demo/url-regex, groups added
+    // group 1: scheme, group 2: domain, group 3: path+query
     private static final Pattern URL_REGEX = Pattern.compile("^(https?)://([^\\s/$.?#].[^\\s/]*)([^\\s]*)$");
 
     private final SessionContextDecorator decorator;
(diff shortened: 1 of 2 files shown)


mridgway (approved):

kyagna (approved):
LGTM

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -374,6 +376,34 @@ private HttpRequestMessage buildZuulHttpRequest(
...
+        // relative uri
+        if (uri.startsWith("/")) {
+            path = uri;
+        } else {
+            Matcher m = URL_REGEX.matcher(uri);
+
+            // absolute uri
+            if (m.matches()) {
+                String match = m.group(3);
+                path = match == null ? uri : match;
```
argha-c, line 390: This doesn't read great. Consider making the assignment explicit on each branch

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -374,6 +376,34 @@ private HttpRequestMessage buildZuulHttpRequest(
...
+        String path;
+        // relative uri
+        if (uri.startsWith("/")) {
+            path = uri;
+        } else {
+            Matcher m = URL_REGEX.matcher(uri);
+
+            // absolute uri
+            if (m.matches()) {
+                String match = m.group(3);
```
argha-c, line 389: Consider adding an example URI in the comment. `3` reads like a magic number.

argha-c (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
artgon: Fixed.

zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
artgon: Added comment at the top.

PR #1194, opened 2022-02-03
ServerTest verifies the number of NioSocketChannel's

sullis:


sullis:
wdyt? @argha-c

zuul-core/src/test/java/com/netflix/zuul/netty/server/ServerTest.java
```diff
@@ -68,9 +71,20 @@ public void beforeTest() {
     public void getListeningSockets() throws Exception {
         ServerStatusManager ssm = mock(ServerStatusManager.class);
         Map<NamedSocketAddress, ChannelInitializer<?>> initializers = new HashMap<>();
+        final List<NioSocketChannel> nioChannels = Collections.synchronizedList(new ArrayList<NioSocketChannel>());
```
argha-c, line 74: Why is this specific to NIO channels?

sullis: > Why is this specific to NIO channels?

ServerTest uses **force_nio**  true

```
config.setProperty("zuul.server.netty.socket.force_nio", "true");
```

so the channels should be NioSocketChannel objects

argha-c: Ah, I forgot we updated that in the last change.

argha-c (approved):

PR #1214, opened 2022-02-27
restore io_uring unit test

sullis:
Let's try using 'sudo -E' in the snapshot workflow

# Reference

https://man7.org/linux/man-pages/man8/sudo.8.html

```
-E

Indicates to the security policy that the user wishes 
to preserve their existing environment variables.
```

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -19,7 +19,7 @@ jobs:
           java-version: 8
           cache: 'gradle'
       - name: Build snapshot
-        run: ./gradlew build snapshot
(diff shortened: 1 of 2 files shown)


sullis:
> @argha-c I don't have a good way to test this. Can you review it?

WDYT?

sullis:
@argha-c  WDYT?

argha-c:
> @argha-c WDYT?

I think we need to a simple way to verify that failing to set the ulimit does not break the snapshot publishing.
As an alternative, we could consider running that test under a separate stage. No strong opinions as long as we confirm the snapshot publish is intact.

PR #1222, opened 2022-03-04
Added category to Filter to allow category metadata assignment of filters

gavinbunney:
Added category field to Filter, to allow specifying of different category metadata for filters.


zuul-core/src/main/java/com/netflix/zuul/Filter.java
```diff
@@ -46,6 +46,11 @@
      */
     FilterType type() default FilterType.INBOUND;
 
+    /**
+     * Category of the filter.
+     */
+    String category() default "zuul";
```
argha-c, line 52: This feels too open-ended as a string representation. 
At the very least, it should have a pre-defined set of categories in an enum etc.

PR #1266, opened 2022-06-22
[all] upgrade to Netty 4.1.78

sullis:
https://netty.io/news/2022/06/14/4-1-78-Final.html

Note:  this PR changes the Gradle dependencies in zuul-core

We needed to explicitly declare platform specific classifiers for the tcnative boringssl library.

This was motivated by an internal change in the Netty project:

```
The maven packaging changed a bit in version 2.0.49.Final.
The changes were fine for Maven users, but causes some 
trouble for Gradle and Bazel users. For Gradle, the solution 
is to explicitly declare the tcnative dependencies with classifiers.
```

source: https://netty.io/wiki/forked-tomcat-native.html


sullis:
fyi  @normanmaurer

zuul-core/src/test/java/com/netflix/zuul/netty/ssl/SslTest.java
```diff
@@ -0,0 +1,43 @@
...
+import io.netty.handler.ssl.SslProvider;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(JUnit4.class)
+public class SslTest {
```
argha-c, line 29: These tests should probably be in a separate PR. 
Consider moving them as part of unit tests for `BaseSslContextFactory`.

argha-c (approved):

zuul-core/src/test/java/com/netflix/zuul/netty/ssl/SslTest.java
sullis: Agreed.  I'll move the unit test(s) to a separate PR.

PR #1281, opened 2022-08-03
Fix retrying requests with body contents by resetting the reader index of ByteBuf

gavinbunney:
For TLS handled connections, once the request body has been written to the channel, the ByteBuf.readerIndex is mutated to point to the end of the buffer. This means when retrying requests, subsequent fetches of that buffer return no bytes (as they use the readerIndex). This only affects TLS connections as the SSL handler in netty mutates those indexes (doesn't appear to be the case with plaintext).

Original issue raised with netty from 2018 (https://github.com/netty/netty/issues/7753) eluded to this fact before the changes to pre-buffer bodies was put into zuul (i.e. pre-buffering on 503s that was introduced hid this _actual issue_ of the indexes needing to be reset).

This change resets the readerIndex before retrying a request, and removes the buffering of bodies in retry scenarios.


zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
```diff
@@ -178,6 +178,13 @@ public Iterable<HttpContent> getBodyContents() {
         return Collections.unmodifiableList(bodyChunks);
     }
 
+    @Override
+    public void resetBodyReader() {
+        for (final HttpContent chunk : bodyChunks) {
```
argha-c, line 183: Is it worth only resetting this when we know we have a body, i.e. effectively `setBody` has been called?

argha-c (approved):

zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
gavinbunney: Looking at the _soft contract_ we don't set the `hasBody` boolean in the `setBody` method. So going to not check that here as the hasBody doesn't determine any logic within the message impl currently.

artgon:
LGTM! This is a huge improvement over the hack we put in back in 2018.

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -896,6 +846,10 @@ protected void handleOriginNonSuccessResponse(final HttpResponse originResponse,
                     startedSendingResponseToClient, zuulRequest.hasCompleteBody(), zuulRequest.getMethod());
             //detach from current origin.
             unlinkFromOrigin();
+
+            // ensure body reader indexes are reset so retry is able to access the body buffer
```
artgon, line 850: Worth clarifying in this comment that we want Netty to have access to that ByteBuf, rather than Zuul itself.

artgon (approved):

PR #1282, opened 2022-08-09
Do not propagate exception once context has been closed in PushRegistrationHandler

AlexanderEllis:
Because PushRegistrationHandler handles the exception and closes the context, it does not need to pass the exception on past itself.

Also removed the `final` so it can be overridden.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushRegistrationHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushRegistrationHandler.java
@@ -95,10 +95,9 @@ public class PushRegistrationHandler extends ChannelInboundHandlerAdapter {
     }
 
     @Override
-    public final void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         logger.error("Exception caught, closing push channel for " + authEvent, cause);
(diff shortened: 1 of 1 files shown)


artgon:
LGTM

zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushRegistrationHandler.java
```diff
@@ -95,10 +95,9 @@ public final void channelInactive(ChannelHandlerContext ctx) throws Exception {
     }
 
     @Override
-    public final void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         logger.error("Exception caught, closing push channel for " + authEvent, cause);
         ctx.close();
```
gavinbunney, line 100: Do we also need to `closeOnFlush` here?

```
if (ch.isActive()) {
    ch.writeAndFlush(Unpooled.EMPTY_BUFFER).addListener(ChannelFutureListener.CLOSE);
}
```

AlexanderEllis: Synced offline — we can keep this change small and keep the `close()` behavior while still removing the `final` in case we (or anyone else) wants to override with the fancier `closeOnFlush` 👍

PR #1283, opened 2022-08-11
Move rate-limited check and add release in PushMessageSender to avoid leak

AlexanderEllis:
When running this with `io.netty.leakDetection.level=paranoid`, there was a leak related to the PushMessageSender `channelRead0`:

```
2022-08-09T14:11:49.764 ERROR [Salamander-ClientToZuulWorker-0] io.netty.util.ResourceLeakDetector: LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
#1:
...
#2:
	io.netty.buffer.AdvancedLeakAwareCompositeByteBuf.retain(AdvancedLeakAwareCompositeByteBuf.java:36)
	com.netflix.zuul.netty.server.push.PushMessageSender.channelRead0(PushMessageSender.java:141)
	com.netflix.zuul.netty.server.push.PushMessageSender.channelRead0(PushMessageSender.java:55)
        ...
```

After adding log lines to surface the refCnt and some basic testing,  I found that `refCnt` goes to 0 after the `pushConn.sendPushMessage` `clientFuture` listener is called, so the additional reference count from the `retain` must be getting released as part of the `ctx.channel().writeAndFlush` in the WS PushProtocol. But, if `sendPushMessage` is skipped due to rate limiting or no readable bytes, the extra count is never released.

This PR does two things:
1. Because the rate limit check doesn't depend on the body, we can have that before the body check (so we don't have to worry about releasing)
2. If there aren't readable bytes, we can release the body before returning. This matches up well with the Netty guidelines:

> If a [sending] component is supposed to pass a reference-counted object to another [receiving] component, the sending component usually does not need to destroy it but defers that decision to the receiving component.

In this case, if we don't end up sending it, we don't need the extra reference count.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushMessageSender.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushMessageSender.java
@@ -38,8 +38,10 @@ import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpUtil;
+
 import javax.inject.Inject;
 import javax.inject.Singleton;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushMessageSender.java
```diff
@@ -138,18 +140,21 @@ protected void channelRead0(final ChannelHandlerContext ctx, final FullHttpReque
...
                 logRateLimited();
                 return;
             }
 
+            final ByteBuf body = request.content().retain();
+            if (body.readableBytes() <= 0) {
+                sendHttpResponse(ctx, request, NO_CONTENT, userAuth);
+                // Because we are not passing the body to the pushConn (who would normally handle destroying),
+                // we need to release it here.
+                body.release();
```
artgon, line 154: It's better practice to use `ReferenceCountUtil.release()`

AlexanderEllis: Nice! Updated.

artgon (approved):

PR #1284, opened 2022-08-19
Allow less strict host header parsing to handle non-RFC2396 compliant host headers

gavinbunney:
The URI parsing changes done in #747 better handled IPv6 addresses, however it introduces issues when host headers contain non-RFC2396 compliant hostnames (for example any `_` or `^` or url-encoded characters).

This PR adds a less strict validation fallback to split-style parsing of the host header when the URI default parsing fails.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
@@ -20,6 +20,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.netflix.config.CachedDynamicBooleanProperty;
 import com.netflix.config.CachedDynamicIntProperty;
 import com.netflix.config.DynamicStringProperty;
+import com.netflix.util.Pair;
 import com.netflix.zuul.context.CommonContextKeys;
 import com.netflix.zuul.context.SessionContext;
 import com.netflix.zuul.filters.ZuulFilter;
@@ -52,6 +53,10 @@ import org.slf4j.LoggerFactory;
 public class HttpRequestMessageImpl implements HttpRequestMessage
 {
     private static final Logger LOG = LoggerFactory.getLogger(HttpRequestMessageImpl.class);
+
+    private static final CachedDynamicBooleanProperty STRICT_HOST_HEADER_VALIDATION = new CachedDynamicBooleanProperty(
+            "zuul.HttpRequestMessage.host.header.strict.validation", true
+    );
 
     private static final CachedDynamicIntProperty MAX_BODY_SIZE_PROP = new CachedDynamicIntProperty(
             "zuul.HttpRequestMessage.body.max.size", 15 * 1000 * 1024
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -563,6 +557,48 @@ public Optional<Integer> getClientDestinationPort() {
...
+        }
+
+        try {
+            // attempt to use default URI parsing - this can fail when not strictly following RFC2396,
+            // for example, having underscores in host names will fail parsing
+            URI uri = new URI(/* scheme= */ null, host, /* path= */ null, /* query= */ null, /* fragment= */ null);
+            if (uri.getHost() != null) {
+                return new Pair<>(uri.getHost(), uri.getPort());
+            }
+        } catch (URISyntaxException ignored) { }
```
artgon, line 580: Debug logging?

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -563,6 +557,48 @@ public Optional<Integer> getClientDestinationPort() {
...
+
+        // fallback to using a colon split
+        // valid IPv6 addresses would have been handled already so any colon is safely assumed a port separator
```
artgon, line 583: What if it's an invalid IPv6 address?

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -563,6 +557,48 @@ public Optional<Integer> getClientDestinationPort() {
...
+            return new Pair<>(null, -1);
+        }
+
+        String parsedHost = components[0];
+        int parsedPort = -1;
+        if (components.length > 1) {
+            try {
+                parsedPort = Integer.parseInt(components[1]);
+            } catch (NumberFormatException ignored) {
+                // ignore failing to parse port numbers and fallback to default port
```
artgon, line 596: Debug logging?

artgon:
I'm not convinced that we want this to be our default functionality. If we do want to allow it, there can be a feature flag that enables it.

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
gavinbunney: Added 🪨

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
gavinbunney: There is a fallback below that detects more than 1 colon (e.g. a non-bracketed IPv6 address) which will throw an exception

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
gavinbunney: Added 🪨

gavinbunney:
> I'm not convinced that we want this to be our default functionality. If we do want to allow it, there can be a feature flag that enables it.

@artgon Gated this behind a property `zuul.HttpRequestMessage.host.header.strict.validation` which defaults to `true` so it can be selected enabled (or well, disabled)

zuul-core/src/test/java/com/netflix/zuul/message/http/HttpRequestMessageImplTest.java
```diff
@@ -45,6 +49,12 @@
 public class HttpRequestMessageImplTest {
 
     HttpRequestMessageImpl request;
+    private final AbstractConfiguration config = ConfigurationManager.getConfigInstance();
+
+    @After
+    public void resetConfig() {
+        config.clearProperty("zuul.HttpRequestMessage.host.header.strict.validation");
```
artgon, line 56: nice trick!

artgon (approved):

PR #1294, opened 2022-08-30
PushRegistrationHandler holds onto ChannelHandlerContext references

jguerra:
After a client successfully authenticates `PushRegistrationHandler` registers a timer to `requestClientToCloseConnection`. However, if the client connection terminates prematurely (connection reset, etc) the scheduled task does not go away and maintains a reference `ChannelHandlerContext`. This PR adds logic to track scheduled tasks and cancel them when the channel becomes inactive


artgon:
Would be nice to have a test for this class, but otherwise, LGTM.

artgon (approved):

gavinbunney (approved):

jguerra:
> 

Added tests

PR #1313, opened 2022-10-04
Do not close connections on 503 responses

artgon:
This change will treat 503 responses as normal responses, after which connections will be returned to the pool, like any other response. This will help reduce connection churn during massive throttling spikes on the back end.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -803,15 +803,9 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
         ClientException.ErrorType niwsErrorType;
 
         if (respStatus == 503) {
-            //Treat 503 status from Origin similarly to connection failures, ie. we want to back off from this server
             statusCategory = FAILURE_ORIGIN_THROTTLED;
             niwsErrorType = ClientException.ErrorType.SERVER_THROTTLED;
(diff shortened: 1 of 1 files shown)


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -803,15 +803,9 @@ protected void handleOriginNonSuccessResponse(final HttpResponse originResponse,
         ClientException.ErrorType niwsErrorType;
 
         if (respStatus == 503) {
-            //Treat 503 status from Origin similarly to connection failures, ie. we want to back off from this server
             statusCategory = FAILURE_ORIGIN_THROTTLED;
             niwsErrorType = ClientException.ErrorType.SERVER_THROTTLED;
             obe = new OutboundException(OutboundErrorType.SERVICE_UNAVAILABLE, requestAttempts);
-            if (originConn != null) {
-                originConn.getServer().incrementSuccessiveConnectionFailureCount();
-                originConn.getServer().addToFailureCount();
```
gavinbunney, line 812: Think we should still track the failure counter here? Otherwise we are missing the server stats on failures

artgon: I think these particular failure counters are intended to track server errors, e.g., connectivity problems. If we don't want 503 to fall into that category anymore, we probably don't want to avoid these servers in the future.

gavinbunney: 🚀

anirudhmendiratta (approved):

PR #1317, opened 2022-10-12
Explicitly release bytebufs when reads are not propagated

artgon:


sullis (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2ContentLengthEnforcingHandler.java
```diff
@@ -52,39 +53,51 @@ public final class Http2ContentLengthEnforcingHandler extends ChannelInboundHand
...
                     ctx.writeAndFlush(new DefaultHttp2ResetFrame(Http2Error.PROTOCOL_ERROR));
                     return;
                 }
-            }
-            if (hasContentLength() && HttpUtil.isTransferEncodingChunked(req)) {
-                // TODO(carl-mastrangelo): this is not right, but meh.  Fix this to return a proper 400.
-                ctx.writeAndFlush(new DefaultHttp2ResetFrame(Http2Error.PROTOCOL_ERROR));
-                return;
+            } finally {
+                ReferenceCountUtil.safeRelease(msg);
```
gavinbunney, line 76: Since we are firing a channel read on Line 103 - I don't think we want to release in those cases, rather just when hitting the writeAndFlush/return blocks

artgon: Ah yeah, I was trying to minimize the diff but probably need to be inside each if block

sullis (approved):
lgtm

PR #1325, opened 2022-10-12
Add safety callback in case bytebuf is not released

artgon:
Because of the complex interaction of `OriginResponseReceiver` and `ProxyEndpoint`, I added a safety callback to release the ByteBuf in case it ever is unreleased at channel close.


gavinbunney (approved):

sullis (approved):

jguerra (approved):

sullis (approved):

artgon:
After testing, this approach does not work for cases where the backend connection is closed abruptly but we can still proxy the response to the client.

PR #1343, opened 2022-10-17
Release chunks on filter exceptions

artgon:
If we're processing a chunk and a filter exception happens, we bail out of the request and return a 500. Since we're in the middle of handling the chunk and may not have added it to the body contents, we need to make sure to release it.


jguerra (approved):
Nice find

sullis (approved):

PR #1344, opened 2022-10-17
disposeBufferedBody on ZuulEndPointRunner exception

sullis:


sullis:
This change is not necessary.   
Per Slack discussion "the body will get disposed when the channel is closed"

PR #1350, opened 2022-10-20
Prevent stray chunks from reaching filter chain

artgon:
If a client has gone away, but the origin is still sending us content chunks, we should release them as soon as possible.


gavinbunney (approved):
nods enthusiastically

jguerra (approved):

PR #1353, opened 2022-10-20
Ensure response filters are not left waiting for buffered body

artgon:
In case of client cancellations or connectivity errors, we may have a response filter chain hanging, waiting for a complete buffered body. The goal for this PR is to close out that filter chain and clean up any body chunks being processed.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/filter/ZuulFilterChainHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/filter/ZuulFilterChainHandler.java
@@ -34,6 +34,7 @@ import com.netflix.zuul.message.http.HttpRequestMessage;
 import com.netflix.zuul.message.http.HttpResponseMessage;
 import com.netflix.zuul.message.http.HttpResponseMessageImpl;
 import com.netflix.zuul.netty.RequestCancelledEvent;
+import com.netflix.zuul.netty.SpectatorUtils;
 import com.netflix.zuul.stats.status.StatusCategory;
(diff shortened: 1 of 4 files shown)


sullis (approved):

kyagna (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/filter/ZuulFilterChainHandler.java
```diff
@@ -121,15 +122,18 @@ private void sendResponse(final StatusCategory statusCategory, final int status,
             headers.add("Content-Length", "0");
             zuulResponse.finishBufferedBodyIfIncomplete();
             responseFilterChain.filter(zuulResponse);
-            fireEndpointFinish(true);
+            fireEndpointFinish(true, ctx);
```
gavinbunney, line 125: One slight nuance here with the change to `fireEndpointFinish`, since we close out the body on L123 (`zuulResponse.finishBufferedBodyIfIncomplete()`) and fire into the response filter chain, I don't think the new `finishResponseFilters` function will have an issue as no filter should be waiting for a body? Just double checking 😅

gavinbunney (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/filter/ZuulFilterChainHandler.java
artgon: True, but in the case of that static response, it won't pop into the if block that checks if the filter chain is waiting for a buffered body.

PR #1354, opened 2022-10-20
Only touch bytebufs when doing advanced logging

artgon:


zuul-core/src/main/java/com/netflix/netty/common/ByteBufUtil.java
```diff
@@ -32,7 +31,7 @@
 public class ByteBufUtil {
 
     private static final boolean isAdvancedLeakDetection =
-            Lists.newArrayList(PARANOID, ADVANCED).contains(ResourceLeakDetector.getLevel());
+            ResourceLeakDetector.getLevel() == PARANOID || ResourceLeakDetector.getLevel() == ADVANCED;
```
gavinbunney, line 34: Not against this, but _could_ use ordinal:

```suggestion
            ResourceLeakDetector.getLevel().ordinal() >= ResourceLeakDetector.Level.ADVANCED.ordinal();
```

Either or is fine 🗡️

zuul-core/src/main/java/com/netflix/netty/common/ByteBufUtil.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author Arthur Gonigberg
+ * @since October 20, 2022
+ */
+public class ByteBufUtil {
+
+    private static final boolean isAdvancedLeakDetection =
+            ResourceLeakDetector.getLevel() == PARANOID || ResourceLeakDetector.getLevel() == ADVANCED;
+
+    public static void touch(ReferenceCounted byteBuf, String hint, ZuulMessage msg) {
```
gavinbunney, line 36: nit: could use String.format with vargs to allow for future logging of other objects if required

artgon: I was thinking about that too, but String.format has terrible performance.

gavinbunney (approved):

PR #1421, opened 2022-12-21
use LongAdder for filter concurrent count

sullis:
# Summary
- replace AtomicInteger with LongAdder in BaseFilter

# why use LongAdder?

```
This class [LongAdder] is usually preferable to AtomicLong when
multiple threads update a common sum that 
is used for purposes such as collecting statistics, 
not for fine-grained synchronization control. Under
 low update contention, the two classes have similar 
characteristics. But under high contention,
 expected throughput of this class is significantly 
higher, at the expense of higher space consumption.
```

Javadoc: 
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/atomic/LongAdder.html

# LongAdder microbenchmark results

https://github.com/sullis/microbenchmarks-java/blob/main/README.md

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/BaseFilter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/BaseFilter.java
@@ -23,7 +23,7 @@ import com.netflix.zuul.message.ZuulMessage;
 import com.netflix.zuul.netty.SpectatorUtils;
 import io.netty.handler.codec.http.HttpContent;
 
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.LongAdder;
 
 /**
  * Base abstract class for ZuulFilters. The base class defines abstract methods to define:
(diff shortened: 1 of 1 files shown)


gavinbunney (approved):

sullis:
I am moving this PR to DRAFT status.   LongAdder needs more evaluation.

zuul-core/src/main/java/com/netflix/zuul/filters/BaseFilter.java
```diff
@@ -126,15 +126,15 @@ public HttpContent processContentChunk(ZuulMessage zuulMessage, HttpContent chun
     @Override
     public void incrementConcurrency() throws ZuulFilterConcurrencyExceededException {
         final int limit = filterConcurrencyLimit.get();
-        if ((concurrencyProtectEnabled.get()) && (concurrentCount.get() >= limit)) {
+        if ((concurrencyProtectEnabled.get()) && (concurrentCount.longValue() >= limit)) {
```
jguerra, line 129: I think `LongAdder` is a poor fit here since the sum needs to be calculated before every increment

PR #1438, opened 2023-01-24
Ensure we release FullHttpResponse messages

artgon:
In some cases, the response we get from the origin is a `DefaultFullHttpResponse`. We need to release it in case the client has gone away (i.e. request is decoupled from response).


jguerra (approved):
Nice find

argha-c (approved):

sullis (approved):

PR #1475, opened 2023-02-20
brotli4j 1.10.0

sullis:


sullis:
brotli4j 1.10.0 was compiled with Java 17.  The class files are not compatible with Java 8.

PR #1483, opened 2023-02-27
Add graceful shutdown method to ClientChannelManager

jguerra:
Adds a method to gracefully shutdown ClientChannelManager. A graceful shutdown will close any idle connections, and allow any in-flight requests to finish.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
@@ -213,6 +213,19 @@ public class DefaultClientChannelManager implements ClientChannelManager {
         }
     }
 
+    /**
+     * Gracefully shuts down a DefaultClientChannelManager by allowing in-flight requests to finish before closing the connections.
+     * Idle connections in the connection pools are closed, and any connections associated with an in-flight request
+     * will be closed upon trying to return the connection to the pool
+     */
+    @Override
+    public void gracefulShutdown() {
+        LOG.info("Starting a graceful shutdown of {}", clientConfig.getClientName());
+        shuttingDown = true;
+        dynamicServerResolver.shutdown();
+        perServerPools.values().forEach(IConnectionPool::drain);
+    }
+
     @Override
     public boolean release(final PooledConnection conn) {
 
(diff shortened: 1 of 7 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
```diff
@@ -225,6 +238,8 @@ public boolean release(final PooledConnection conn) {
         discoveryResult.incrementNumRequests();
 
         if (shuttingDown) {
+            LOG.debug("[{}] closing connection released during shutdown", conn.getChannel().id());
+            conn.close();
```
jguerra, line 242: The pool loses track of a connection as soon as it is borrowed. Since we're shutting down it seemed simplest to just immediately close in this case vs returning to to the pool and trying to close there

artgon, line 242: An explicit close is not necessary here. The graceful shutdown should handle closing the connections.

jguerra: I agree that would be better. I was working with the existing logic of not returning the connection to the pool during shutdown, but I'm not sure we need to keep doing that

argha-c: Yeah, `close()` is best delegated to the conn. pool itself.

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/PerServerConnectionPool.java
```diff
@@ -398,4 +423,26 @@ private static InetAddress getSelectedHostString(SocketAddress addr) {
...
+    /**
+     * Closes idle connections in the connection pool for a given EventLoop. The closing is performed on the EventLoop
+     * thread since the connection pool is not thread safe.
+     *
+     * @param eventLoop - the event loop to drain
+     */
+    void drainIdleConnectionsOnEventLoop(EventLoop eventLoop) {
+        eventLoop.execute(() -> {
+            Deque<PooledConnection> connections = connectionsPerEventLoop.get(eventLoop);
+            for(Iterator<PooledConnection> it = connections.iterator(); it.hasNext(); ) {
```
argha-c, line 435: nit: Any reason to use a iterator here, vs a for-each? 
For this case, it should  just use a iterator underneath, if performance vs readability is the tradeoff.

jguerra: I used an iterator so I could use remove(). However, I don't need to be removing the connection from the backing Deque (it was required on an earlier implementation I had). I can switch to a for-each

argha-c (approved):

PR #1487, opened 2023-03-01
Replace PromiseCombiner usage on ClientConnectionsShutdown

jguerra:
ClientConnectionsShutdown is broken in two different ways

1) When `gracefullyShutdownClientChannels()` is called as part of a discovery callback, the use of `ImmediateEventExecutor.INSTANCE` triggers a deadlock check in netty:

```
Caused by: io.netty.util.concurrent.BlockingOperationException: DefaultPromise@7ff6ea19(incomplete)
	at app//io.netty.util.concurrent.DefaultPromise.checkDeadLock(DefaultPromise.java:463)
	at app//io.netty.util.concurrent.DefaultPromise.await0(DefaultPromise.java:687)
	at app//io.netty.util.concurrent.DefaultPromise.await(DefaultPromise.java:295)
```

2) The use of PromiseCombiner isn't  thread safe and the future isn't always finished correctly

I replaced the use of PromiseCombiner with bulk operations on the ChannelGroup, and also added unit tests

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientConnectionsShutdown.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientConnectionsShutdown.java
@@ -110,7 +110,6 @@ public class ClientConnectionsShutdown {
         closeFuture.addListener(future -> {
             if (!timeoutTask.isDone()) {
                 //close happened before the timeout
-                LOG.info("All connections closed within timeout");
                 timeoutTask.cancel(false);
             }
             promise.setSuccess(null);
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientConnectionsShutdown.java
```diff
@@ -23,134 +23,99 @@
...
+
+        ScheduledFuture<?> timeoutTask = executor.schedule(() -> {
+            LOG.warn("Force closing remaining {} active client channels.", channels.size());
+            channels.close();
+        }, GRACEFUL_CLOSE_TIMEOUT.get(), TimeUnit.SECONDS);
+
+        closeFuture.addListener(future -> {
+            if (!timeoutTask.isDone()) {
+                //close happened before the timeout
+                LOG.info("All connections closed within timeout");
```
gavinbunney, line 113: nit: do we care? 😉

gavinbunney (approved):

PR #1500, opened 2023-03-11
store Netty Http Response in SessionContext

sullis:


sullis:
after discussing this PR with Arthur, I found an alternate solution. This PR is not necessary.

PR #1531, opened 2023-04-08
Ensure we release FullHttpContent instances when unlinking from origin request

artgon:


sullis (approved):
lgtm

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -721,6 +727,7 @@ public void responseFromOrigin(final HttpResponse originResponse) {
             methodBinding.bind(() -> processResponseFromOrigin(originResponse));
         } catch (Exception ex) {
             unlinkFromOrigin();
+            releasePartialResponse(originResponse);
```
argha-c, line 730: should be able to repro this in a unit test?

argha-c (approved):

PR #1540, opened 2023-04-17
suppress deep client side ssl handshake stacktraces

argha-c:
In the absence of this today, we let these propagate through to the filter chain handler. The logging there is necessary, but happens at a higher levels.
This preserves the filter chain logging, but suppresses this specific class of exception from propagating up the pipeline.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/netty/common/SslExceptionsHandler.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/SslExceptionsHandler.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2023 Netflix, Inc.
+ *
+ *      Licensed under the Apache License, Version 2.0 (the "License");
+ *      you may not use this file except in compliance with the License.
+ *      You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *      Unless required by applicable law or agreed to in writing, software
+ *      distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/netty/common/SslExceptionsHandler.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        // In certain cases, depending on the client, these stack traces can get very deep.
+        // We intentionally avoid propagating this up the pipeline, to avoid verbose disk logging.
+        if (cause.getCause() instanceof SSLHandshakeException) {
+            logger.debug("SSL handshake failed on channel {}", ctx.channel(), cause);
+            registry.counter("server.ssl.exception.swallowed", "cause", "SSLHandshakeException");
+        }
```
jguerra, line 34: For other exception types don't we still want to propagate?

sullis:
CI build failed:
```
Missing header in: zuul-core/src/main/java/com/netflix/netty/common/SslExceptionsHandler.java
```

zuul-core/src/main/java/com/netflix/netty/common/SslExceptionsHandler.java
argha-c: Yes, hence the specific check. The changeset was incomplete, should be updated now.

jguerra (approved):

sullis (approved):

PR #1581, opened 2023-05-16
Allow injecting shutdown hooks. Defer thread creation in Singletons

jguerra:


zuul-core/src/main/java/com/netflix/zuul/FilterFileManager.java
```diff
@@ -72,10 +69,16 @@ public FilterFileManager(FilterFileManagerConfig config, FilterLoader filterLoad
      * @throws Exception
      */
     @Inject
```
argha-c, line 71: Shouldn't this be a`@PostConstruct` vs `@Inject`?

argha-c (approved):

argha-c (approved):

PR #1638, opened 2023-09-08
Properly apply socket options

jguerra:
These errors have always annoyed me:

```
2023-09-08T15:30:58.748 WARN  [main] io.netty.bootstrap.ServerBootstrap: Unknown channel option 'SO_LINGER' for channel '[id: 0x08c9bd45]'
2023-09-08T15:30:58.748 WARN  [main] io.netty.bootstrap.ServerBootstrap: Unknown channel option 'SO_KEEPALIVE' for channel '[id: 0x08c9bd45]'
2023-09-08T15:30:58.748 WARN  [main] io.netty.bootstrap.ServerBootstrap: Unknown channel option 'TCP_NODELAY' for channel '[id: 0x08c9bd45]'
```

turns out we were attempting to apply the options incorrectly

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
@@ -269,20 +269,14 @@ public class Server
         ServerBootstrap serverBootstrap =
                 new ServerBootstrap().group(serverGroup.clientToProxyBossPool, serverGroup.clientToProxyWorkerPool);
 
-        // Choose socket options.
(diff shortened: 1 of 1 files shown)


gavinbunney (approved):
🤨 Does that mean we haven't had keepalive enabled at all?

jguerra:
> 🤨 Does that mean we haven't had keepalive enabled at all? Looking at the docs, seems like that's the case, but more from server-side, so may cause more server-side closures... neat



> 

Yeah, it has not been enabled. I expect the impact to be fairly minimal because TCP keepalives only occur after ~2 hours, and we have aggressive idle timeouts on all connections.

PR #1667, opened 2023-10-20
Add method to get all entries from push conn registry

kyagna:
Change adds a new API/method to retrieve all entries from the push connection registry. The returned list represents a snapshot of the map as it was when the call was made.


AlexanderEllis (approved):
Nice, thanks for adding a test as well!

PR #1668, opened 2023-10-20
Separate adding push handlers in initializer

kyagna:
Add new method to handle adding of push handlers. This allows extending the method to add any additional handlers.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushMessageSenderInitializer.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushMessageSenderInitializer.java
@@ -26,20 +26,13 @@ import io.netty.handler.codec.http.HttpServerCodec;
  * Date: 5/16/18
  */
 public abstract class PushMessageSenderInitializer extends ChannelInitializer<Channel> {
-
-    private final PushConnectionRegistry pushConnectionRegistry;
-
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/push/PushMessageSenderInitializer.java
```diff
@@ -26,20 +26,13 @@
...
     protected void initChannel(Channel ch) throws Exception {
         final ChannelPipeline pipeline = ch.pipeline();
         pipeline.addLast(new HttpServerCodec());
         pipeline.addLast(new HttpObjectAggregator(65536));
-        pipeline.addLast(getPushMessageSender(pushConnectionRegistry));
+        addPushMessageHandlers(pipeline);
     }
 
-    protected abstract PushMessageSender getPushMessageSender(PushConnectionRegistry pushConnectionRegistry);
+    protected abstract void addPushMessageHandlers(final ChannelPipeline pipeline);
```
AlexanderEllis, line 37: Any concerns about this being a breaking change for folks? Not sure if you'd want to have a separate function to add other handlers instead of replacing, but I'll leave it up to you.

AlexanderEllis (approved):

sullis (approved):

PR #1673, opened 2023-10-22
add NettyTransportTest

sullis:


sullis:
This PR is not needed. We already have assertions in IntegrationTest.java

PR #1683, opened 2023-10-27
spectator 1.7.0

sullis:
https://github.com/Netflix/spectator/releases


zuul-core/build.gradle
```diff
@@ -12,7 +12,7 @@ dependencies {
     api 'com.fasterxml.jackson.core:jackson-databind:2.14.2'
 
     api "com.netflix.archaius:archaius-core:0.7.7"
-    api "com.netflix.spectator:spectator-api:1.6.11"
+    api "com.netflix.spectator:spectator-api:1.7.0"
```
gavinbunney, line 15: Think we can just pin to latest release?

PR #1701, opened 2023-11-15
Add a new handler to close on inbound http/2 protocol errors 

jguerra:
Handles ~closing the connection~ logging/counting malformed requests like invalid hpack, or invalid stream numbers.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2ProtocolErrorHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2ProtocolErrorHandler.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2ProtocolErrorHandler.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author Justin Guerra
+ * @since 11/14/23
+ */
+public class Http2ProtocolErrorHandler extends ChannelInboundHandlerAdapter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(Http2ProtocolErrorHandler.class);
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        if(cause instanceof Http2Exception http2Exception && http2Exception.shutdownHint() == Http2Exception.ShutdownHint.HARD_SHUTDOWN) {
```
argha-c, line 40: The `Http2ConnectionHandler` has a generic mechanism to send GO_AWAY frames and close on connection errors. 
Is that not kicking in here or are we trying to address specific cases that aren't flagged via that mechanism?

argha-c (approved):

PR #1704, opened 2023-11-17
Add empty fn for doing things with the client and origin channels

AlexanderEllis:
Just a nice empty function for now to allow for overriding and experimentation.


argha-c:
Can you explain the issue this solves more?
The comment gives a hint but I want to understand the requirements better.

AlexanderEllis:
@argha-c  the general idea is to enable some gentle experimentation with some sort of synchronization between the origin and client channel for servers with unique request patterns.  Namely, cases where the origin channel and the client channel being out of band leads to issues.

Just to emphasize again, this is not intended to be a large-scale change to zuul's behavior. This is just adding a hook to let us play with this idea more.

PR #1705, opened 2023-11-23
Make channelReadInternal protected to allow for override

AlexanderEllis:
Mostly enabling experimentation around these origin channel read calls.


AlexanderEllis:
Super minor change, so going to 🤠

PR #1736, opened 2024-01-30
Only capture first line of stacktrace for unexpected RequestAttempt exceptions

gavinbunney:


zuul-core/src/main/java/com/netflix/zuul/niws/RequestAttempt.java
```diff
@@ -340,7 +339,13 @@ public void setException(Throwable t) {
             } else {
                 error = t.getMessage();
                 exceptionType = t.getClass().getSimpleName();
-                cause = Throwables.getStackTraceAsString(t);
+
+                // for unexpected exceptions, just capture the first line of the stacktrace
```
argha-c, line 343: If we're suppressing the trace, doesn't  `cause.getCause() != null ? cause.getCause().toString() : cause.toString())` give you the hint?
When we do suppress, consider making it conditional instead.

PR #1749, opened 2024-02-27
Switch to io.netty.handler.codec.http.cookie.Cookie in order to support cookies with duplicate keys

jguerra:
We've been using the deprecated `io.netty.handler.codec.http.Cookie` and `CookieDecoder.decode`. `CookieDecoder` is unable to handle duplicate key names, so I had to switch us to the newer decoder. I'm also bumping to 2.5.0 since I had to make breaking changes to interfaces

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,5 +4,5 @@ versions_netty=4.1.107.Final
 versions_netty_io_uring=0.0.25.Final
 versions_brotli4j=1.16.0
 release.scope=patch
-release.version=2.5.1-SNAPSHOT
+release.version=2.5.0-SNAPSHOT
 org.gradle.jvmargs=-Xms1g -Xmx2g
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,5 +4,5 @@ versions_netty=4.1.107.Final
 versions_netty_io_uring=0.0.25.Final
 versions_brotli4j=1.16.0
 release.scope=patch
-release.version=2.4.1-SNAPSHOT
+release.version=2.5.1-SNAPSHOT
 org.gradle.jvmargs=-Xms1g -Xmx2g
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
@@ -16,7 +16,7 @@
 
 package com.netflix.zuul.message.http;
 
-import io.netty.handler.codec.http.Cookie;
+import io.netty.handler.codec.http.cookie.Cookie;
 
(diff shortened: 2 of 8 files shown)


gradle.properties
```diff
@@ -4,5 +4,5 @@ versions_netty=4.1.107.Final
 versions_netty_io_uring=0.0.25.Final
 versions_brotli4j=1.16.0
 release.scope=patch
-release.version=2.4.1-SNAPSHOT
+release.version=2.5.1-SNAPSHOT
```
gavinbunney, line 7:
```suggestion
release.version=2.5.0-SNAPSHOT
```

zuul-core/src/main/java/com/netflix/zuul/message/http/Cookies.java
```diff
@@ -29,16 +29,11 @@
...
+    private final List<Cookie> all = new ArrayList<>();
 
     public void add(Cookie cookie) {
-        List<Cookie> existing = map.get(cookie.getName());
-        if (existing == null) {
-            existing = new ArrayList<>();
-            map.put(cookie.getName(), existing);
-        }
-        existing.add(cookie);
+        map.computeIfAbsent(cookie.name(), k -> new ArrayList<>(1)).add(cookie);
```
gavinbunney, line 36: 💯

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpResponseMessageImpl.java
```diff
@@ -179,24 +179,19 @@ public int getMaxBodySize() {
...
     @Override
     public boolean hasSetCookieWithName(String cookieName) {
-        boolean has = false;
         for (String setCookieValue : getHeaders().getAll(HttpHeaderNames.SET_COOKIE)) {
-            for (Cookie cookie : CookieDecoder.decode(setCookieValue)) {
-                if (cookie.getName().equalsIgnoreCase(cookieName)) {
-                    has = true;
-                    break;
-                }
+            Cookie cookie = ClientCookieDecoder.STRICT.decode(setCookieValue);
```
gavinbunney, line 189: Intentional to use STRICT here vs LAX in `HttpRequestMessageImpl.java` and further down in add/setCookie?

jguerra: Yeah, I maintained the strict/lax behavior. By default CookieDecoder.decode() uses strict, but for whatever reason ServerCookieEncoder.encode uses lax

gavinbunney (approved):

PR #1766, opened 2024-04-19
Add Lazy cache for originalhost

karim-z:
Add Lazy cache for originalhost when parsing headers

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
@@ -64,9 +64,6 @@ public class HttpRequestMessageImpl implements HttpRequestMessage {
     private static final DynamicStringProperty REGEX_PTNS_TO_STRIP_PROP =
             new DynamicStringProperty("zuul.request.cookie.cleaner.strip", " Secure,");
 
-    private static final CachedDynamicBooleanProperty CACHE_ORIGINAL_HOST =
-            new CachedDynamicBooleanProperty("zuul.HttpRequestMessage.original.host.cache", false);
-
     private static final List<Pattern> RE_STRIP;
 
     static {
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -510,12 +514,23 @@ protected String generateInfoForLogging() {
     @Override
     public String getOriginalHost() {
         try {
-            return getOriginalHost(getHeaders(), getServerName());
+            if (!CACHE_ORIGINAL_HOST.get()) {
+                return generateOriginalHost(getHeaders(), getServerName());
```
gavinbunney, line 518: I'm not sure we need this FP to not-cache the original host; since it's intended for fetching the host on the inbound request, it shouldn't be mutated by any header processing, so keeping it always cached makes it simpler

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -510,12 +514,23 @@ protected String generateInfoForLogging() {
...
+            if (originalHost == null) {
+                originalHost = generateOriginalHost(getHeaders(), getServerName());
+            }
+            return originalHost;
         } catch (URISyntaxException e) {
             throw new IllegalArgumentException(e);
         }
     }
 
+    String generateOriginalHost(Headers headers, String serverName) throws URISyntaxException {
```
gavinbunney, line 530: Think you can remove this method and just call `getOriginalHost(headers, serverName)` from within the `getOriginalHost()` method

zuul-core/src/main/java/com/netflix/zuul/message/http/HttpRequestMessageImpl.java
```diff
@@ -510,7 +511,10 @@ protected String generateInfoForLogging() {
     @Override
     public String getOriginalHost() {
         try {
-            return getOriginalHost(getHeaders(), getServerName());
+            if (originalHost == null) {
+                originalHost = getOriginalHost(getHeaders(), getServerName());
```
argha-c, line 515: The compute intensive part here should be the `parseHostHeader` logic. 
Consider the option to just cache that part and not the `X-Forwarded-Host` part.

karim-z: `X-Forwarded-Host` part is called everytime before the `parseHostHeader` and it loops on all existing headers to check for the `x-forwarded-host` using `headers.getFirst` before attempting to call the `parseHostHeader`. Doesn't sound efficient. I am tempted to cache that as well.

argha-c: hmm, it does iterate through the list, ugh.

argha-c (approved):

PR #1773, opened 2024-05-08
Adjust retry policy before checking isBelowRetryLimit

jguerra:


gavinbunney (approved):

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -673,10 +672,14 @@ private void processErrorFromOrigin(final Throwable ex, final Channel origCh) {
                 origin.onRequestExceptionWithServer(zuulRequest, chosenServer.get(), attemptNum, niwsEx);
             }
 
-            if ((isBelowRetryLimit()) && (isRetryable(err))) {
+            boolean retryable = isRetryable(err);
+            if (retryable) {
+                origin.adjustRetryPolicyIfNeeded(zuulRequest);
```
argha-c, line 677: The adjustment itself is fine. 
The issue here is currently this is going to throw off retries disabled metrics, regardless of whether or not the limit was reached. We need to split that part from the adjustment, as part of this change since it's inaccurate otherwise.

argha-c (changes requested):

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
jguerra: Ah good point

jguerra: Ah good call out

PR #1777, opened 2024-05-21
TLS PSK implementation

sunnysingh85:
This PR adds an implementation to use a pre-shared key (PSK) to complete a TLS handshake ([TLS-PSK](https://www.rfc-editor.org/rfc/rfc8446.html#section-2.2)). This implementation is based on PSK APIs offered by BouncyCastle.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/netty/common/ssl/SslHandshakeInfo.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/ssl/SslHandshakeInfo.java
@@ -16,6 +16,7 @@
 
 package com.netflix.netty.common.ssl;
 
+import com.netflix.zuul.netty.server.psk.ClientPSKIdentityInfo;
 import io.netty.handler.ssl.ClientAuth;
 import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
@@ -31,6 +32,8 @@ public class SslHandshakeInfo {
     private final Certificate serverCertificate;
     private final X509Certificate clientCertificate;
     private final boolean isOfIntermediary;
+    private final boolean usingExternalPSK;
+    private final ClientPSKIdentityInfo clientPSKIdentityInfo;
 
     public SslHandshakeInfo(
             boolean isOfIntermediary,
@@ -38,13 +41,17 @@ public class SslHandshakeInfo {
             String cipherSuite,
             ClientAuth clientAuthRequirement,
             Certificate serverCertificate,
-            X509Certificate clientCertificate) {
+            X509Certificate clientCertificate,
+            boolean usingExternalPSK,
+            ClientPSKIdentityInfo clientPSKIdentityInfo) {
         this.protocol = protocol;
         this.cipherSuite = cipherSuite;
         this.clientAuthRequirement = clientAuthRequirement;
         this.serverCertificate = serverCertificate;
         this.clientCertificate = clientCertificate;
         this.isOfIntermediary = isOfIntermediary;
+        this.usingExternalPSK = usingExternalPSK;
+        this.clientPSKIdentityInfo = clientPSKIdentityInfo;
     }
 
     public boolean isOfIntermediary() {
@@ -69,6 +76,14 @@ public class SslHandshakeInfo {
 
     public X509Certificate getClientCertificate() {
         return clientCertificate;
+    }
+
+    public boolean usingExternalPSK() {
+        return usingExternalPSK;
+    }
+
+    public ClientPSKIdentityInfo geClientPSKIdentityInfo() {
+        return clientPSKIdentityInfo;
     }
 
     @Override
--- a/zuul-core/src/main/java/com/netflix/netty/common/ssl/SslHandshakeInfo.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/ssl/SslHandshakeInfo.java
@@ -35,6 +35,24 @@ public class SslHandshakeInfo {
     private final boolean usingExternalPSK;
     private final ClientPSKIdentityInfo clientPSKIdentityInfo;
 
+    //for backward compatibility
+    public SslHandshakeInfo(
+            boolean isOfIntermediary,
+            String protocol,
+            String cipherSuite,
+            ClientAuth clientAuthRequirement,
+            Certificate serverCertificate,
+            X509Certificate clientCertificate) {
+        this.protocol = protocol;
+        this.cipherSuite = cipherSuite;
+        this.clientAuthRequirement = clientAuthRequirement;
+        this.serverCertificate = serverCertificate;
+        this.clientCertificate = clientCertificate;
+        this.isOfIntermediary = isOfIntermediary;
+        this.usingExternalPSK = false;
+        this.clientPSKIdentityInfo = null;
+    }
+
     public SslHandshakeInfo(
             boolean isOfIntermediary,
             String protocol,
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2OrHttpHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2OrHttpHandler.java
@@ -36,12 +36,11 @@ import io.netty.handler.ssl.ApplicationProtocolNames;
 import io.netty.handler.ssl.ApplicationProtocolNegotiationHandler;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
 import io.netty.util.AttributeKey;
-
 import java.util.function.Consumer;
 
 /**
  * Http2 Or Http Handler
- *
+ * <p>
  * Author: Arthur Gonigberg
  * Date: December 15, 2017
  */
@@ -75,6 +74,9 @@ public class Http2OrHttpHandler extends ApplicationProtocolNegotiationHandler {
         this.addHttpHandlerFn = addHttpHandlerFn;
     }
 
+    /**
+     * this method is inspired by ApplicationProtocolNegotiationHandler.userEventTriggered
+     */
     @Override
     public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
         if (evt instanceof SslHandshakeCompletionEvent handshakeEvent) {
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2OrHttpHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2OrHttpHandler.java
@@ -20,6 +20,7 @@ import com.netflix.netty.common.channel.config.ChannelConfig;
 import com.netflix.netty.common.channel.config.CommonChannelConfigKeys;
 import com.netflix.netty.common.http2.DynamicHttp2FrameLogger;
 import com.netflix.zuul.netty.server.BaseZuulChannelInitializer;
+import com.netflix.zuul.netty.server.psk.TlsPskHandler;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
@@ -33,7 +34,9 @@ import io.netty.handler.codec.http2.UniformStreamByteDistributor;
 import io.netty.handler.logging.LogLevel;
 import io.netty.handler.ssl.ApplicationProtocolNames;
 import io.netty.handler.ssl.ApplicationProtocolNegotiationHandler;
+import io.netty.handler.ssl.SslHandshakeCompletionEvent;
 import io.netty.util.AttributeKey;
+
 import java.util.function.Consumer;
 
 /**
@@ -44,6 +47,8 @@ import java.util.function.Consumer;
  */
 public class Http2OrHttpHandler extends ApplicationProtocolNegotiationHandler {
     public static final AttributeKey<String> PROTOCOL_NAME = AttributeKey.valueOf("protocol_name");
+
+    private static final String FALLBACK_APPLICATION_PROTOCOL = ApplicationProtocolNames.HTTP_1_1;
 
     private static final DynamicHttp2FrameLogger FRAME_LOGGER =
             new DynamicHttp2FrameLogger(LogLevel.DEBUG, Http2FrameCodec.class);
@@ -60,7 +65,7 @@ public class Http2OrHttpHandler extends ApplicationProtocolNegotiationHandler {
             ChannelHandler http2StreamHandler,
             ChannelConfig channelConfig,
             Consumer<ChannelPipeline> addHttpHandlerFn) {
-        super(ApplicationProtocolNames.HTTP_1_1);
+        super(FALLBACK_APPLICATION_PROTOCOL);
         this.http2StreamHandler = http2StreamHandler;
         this.maxConcurrentStreams = channelConfig.get(CommonChannelConfigKeys.maxConcurrentStreams);
         this.initialWindowSize = channelConfig.get(CommonChannelConfigKeys.initialWindowSize);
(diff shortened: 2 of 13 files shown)


padolph (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+        } catch (TlsFatalAlert tlsFatalAlert) {
+            writeOutputIfAvailable(ctx);
+            return;
+        }
+        writeOutputIfAvailable(ctx);
+        final int appDataAvailable = tlsPskServerProtocol.getAvailableInputBytes();
+        if (appDataAvailable > 0) {
+            byte[] appData = new byte[appDataAvailable];
+            tlsPskServerProtocol.readInput(appData, 0, appDataAvailable);
+            out.add(Unpooled.wrappedBuffer(appData));
```
argha-c, line 166: Do we need to pass this data down to other handlers?

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+                }
+
+                @Override
+                public Principal getLocalPrincipal() {
+                    return null;
+                }
+
+                @Override
+                public String getCipherSuite() {
+                    return SUPPORTED_TLS_PSK_CIPHER_SUITE_MAP.get(
```
argha-c, line 468: Is the suite map the accurate value here? 
Or should it be derived from the underlying session, i.e. `SessionParameters` in `this.getContext().getSession()`?

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+        @Override
+        public TlsCredentials getCredentials() {
+            return null;
+        }
+
+        @Override
+        protected Vector getProtocolNames() {
+            Vector protocolNames = new Vector();
+            protocolNames.addElement(ProtocolName.HTTP_1_1);
+            protocolNames.addElement(ProtocolName.HTTP_2_TLS);
```
argha-c, line 268: This list should probably be injected, esp. for clients we don't want to advertise h2 for.

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+        @SneakyThrows // TODO: Ask BC folks to see if getExternalPSK can throw a checked exception
+        public TlsPSKExternal getExternalPSK(Vector clientPskIdentities) {
+            byte[] clientPskIdentity = ((PskIdentity)clientPskIdentities.get(0)).getIdentity();
+            byte[] psk;
+            try{
+                this.ctx.channel().attr(CLIENT_PSK_IDENTITY_ATTRIBUTE_KEY).set(new ClientPSKIdentityInfo(clientPskIdentity));
+                psk = externalTlsPskProvider.provide(clientPskIdentity, this.context.getSecurityParametersHandshake().getClientRandom());
+            }catch (PskCreationFailureException e) {
+                throw switch (e.getTlsAlertMessage()) {
+                    case unknown_psk_identity -> new TlsFatalAlert(AlertDescription.unknown_psk_identity, "Unknown or null client PSk identity");
```
argha-c, line 317: It's unclear from the API if this fails the handshake. Can you clarify how this aborts the handshake? 
We probably need to handle things like firing `SslCloseCompletionEvent` and cleanup manually anyway?

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+                        CipherSuite.TLS_AES_128_GCM_SHA256,
+                        CipherSuite.TLS_CHACHA20_POLY1305_SHA256 -> PRFAlgorithm.tls13_hkdf_sha256;
+                case CipherSuite.TLS_AES_256_GCM_SHA384 -> PRFAlgorithm.tls13_hkdf_sha384;
+                case CipherSuite.TLS_SM4_CCM_SM3, CipherSuite.TLS_SM4_GCM_SM3 -> PRFAlgorithm.tls13_hkdf_sm3;
+                default -> -1;
+            };
+        }
+    }
+
+    static class TlsPskServerProtocol extends TlsServerProtocol {
```
jguerra, line 393: for improved readability it might be nice to extract these static inner classes

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+            return;
+        }
+        byte[] appDataBytes = byteBufMsg.hasArray() ? byteBufMsg.array() : readDirect(byteBufMsg);
+        ReferenceCountUtil.safeRelease(byteBufMsg);
+        tlsPskServerProtocol.writeApplicationData(appDataBytes, 0, appDataBytes.length);
+        int availableOutputBytes = tlsPskServerProtocol.getAvailableOutputBytes();
+        if (availableOutputBytes != 0) {
+            byte[] outputBytes = new byte[availableOutputBytes];
+            tlsPskServerProtocol.readOutput(outputBytes, 0, availableOutputBytes);
+            ctx.write(Unpooled.wrappedBuffer(outputBytes), promise)
```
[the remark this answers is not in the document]
sunnysingh85: I was using `.writeAndFlush` previously but I ran into some issues during load testing with larger payloads and chunks. I can try to reproduce those and share more details.

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+        ctx.flush();
+    }
+
+    @Override
+    protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+        final byte[] bytesRead = in.hasArray() ? in.array() : readDirect(in);
+        try {
+            tlsPskServerProtocol.offerInput(bytesRead);
+        } catch (TlsFatalAlert tlsFatalAlert) {
+            writeOutputIfAvailable(ctx);
```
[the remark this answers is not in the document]
sunnysingh85: Thanks for pointing this out. I'll update the code.

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
sunnysingh85: Based on my understanding, the fact that TlsPskHandler is a `ByteToMessageDecoder`, adding the decoded output to `List<Object> out` does the job of passing down the data to other handlers. LMK if I'm missing the point that you are making.

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
sunnysingh85: I'm actually getting the negotiated cipher suite from context `getContext().getSecurityParameters().getCipherSuite()` which returns an integer but `SSLSession` requires me to return a name of the suite, the map is actually doing the conversion from int to String.

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+
+        @Override
+        public ProtocolVersion getServerVersion() throws IOException {
+            return super.getServerVersion();
+        }
+
+        @Override
+        @SneakyThrows // TODO: Ask BC folks to see if getExternalPSK can throw a checked exception
+        public TlsPSKExternal getExternalPSK(Vector clientPskIdentities) {
+            byte[] clientPskIdentity = ((PskIdentity)clientPskIdentities.get(0)).getIdentity();
```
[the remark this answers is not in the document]
sunnysingh85: AFAIK we only plan to support one PSK identity in the hint, so it should be safe. And BC only invokes this method if clientPskIdentities is non empty.

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,499 @@
...
+import java.security.SecureRandom;
+import java.security.cert.Certificate;
+import java.util.Hashtable;
+import java.util.List;
+import java.util.Map;
+import java.util.Vector;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+public class TlsPskHandler extends ByteToMessageDecoder
```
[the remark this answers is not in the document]
sunnysingh85: This made the code so much cleaner, thanks for this comment!

zuul-core/src/main/java/com/netflix/zuul/netty/server/ssl/SslHandshakeInfoHandler.java
```diff
@@ -98,13 +105,23 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                         serverCert = session.getLocalCertificates()[0];
                     }
 
+                    Boolean tlsHandshakeUsingExternalPSK = ctx.channel()
+                            .attr(ZuulPskServer.TLS_HANDSHAKE_USING_EXTERNAL_PSK)
+                            .get();
```
jguerra, line 110: This looks like a NPE risk since the `SslHandshakeInfo` ctor will unbox this to a boolean. Is TLS_HANDSHAKE_USING_EXTERNAL_PSK always going to be set, or should this be set to false in the case of the attribute having not been set on the channel?

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/ZuulPskServer.java
```diff
@@ -0,0 +1,230 @@
...
+                        .toArray());
+    }
+
+    @Override
+    public ProtocolVersion getServerVersion() throws IOException {
+        return super.getServerVersion();
+    }
+
+    @Override
+    @SneakyThrows
```
jguerra, line 146: Are there other checked exceptions that you're not handling?

zuul-core/src/main/java/com/netflix/zuul/netty/server/http2/Http2OrHttpHandler.java
```diff
@@ -70,6 +75,42 @@ public Http2OrHttpHandler(
         this.addHttpHandlerFn = addHttpHandlerFn;
     }
 
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
```
jguerra, line 79: nit: add a javadoc saying this method is inspired by `ApplicationProtocolNegotiationHandler.userEventTriggered`

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskUtils.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.netflix.zuul.netty.server.psk;
+
+import io.netty.buffer.ByteBuf;
+
+class TlsPskUtils {
+    static byte[] readDirect(ByteBuf byteBufMsg) {
+        int length = byteBufMsg.readableBytes();
+        byte[] dest = new byte[length];
+        byteBufMsg.readSlice(length).getBytes(0, dest);
```
jguerra, line 25: nit: I think `byteBufMsg.readBytes(dest)` essentially does the same thing, but is a little easier to parse

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/TlsPskHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+        if (!(msg instanceof ByteBuf byteBufMsg)) {
+            ReferenceCountUtil.safeRelease(msg);
+            promise.setFailure(new IllegalStateException("Failed to write message on the channel. Message is not a ByteBuf"));
+            return;
+        }
+        byte[] appDataBytes = byteBufMsg.hasArray() ? byteBufMsg.array() : TlsPskUtils.readDirect(byteBufMsg);
```
jguerra, line 70: could all of this logic (including the safeRelease) be put in the helper method?

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/ZuulPskServer.java
```diff
@@ -0,0 +1,230 @@
...
+                case decrypt_error -> new TlsFatalAlert(AlertDescription.decrypt_error, "Invalid or expired client PSk identity");
+            };
+        }
+        TlsSecret pskTlsSecret = getCrypto().createSecret(psk);
+        int prfAlgorithm = getPRFAlgorithm13(getSelectedCipherSuite());
+        return new BasicTlsPSKExternal(clientPskIdentity, pskTlsSecret, prfAlgorithm);
+    }
+
+    @Override
+    public void notifyAlertRaised(short alertLevel, short alertDescription, String message, Throwable cause) {
```
jguerra, line 166: There is one other event netty fires that we'd be interested in (if possible): https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslCloseCompletionEvent.java netty will fire this event when the the ssl session is closed. Does bouncy castle trigger any similar alerts or anything when that happens?

deeptiv1991: Justin
In TLSProtocol.java file I see it raises close_notify alert : https://github.com/bcgit/bc-java/blob/main/tls/src/main/java/org/bouncycastle/tls/TlsProtocol.java#L300 which will eventually make it's way to this notifyAlertRaised method. Did you want us to do something specific with that alert?
https://github.com/bcgit/bc-java/blob/main/tls/src/main/java/org/bouncycastle/tls/TlsProtocol.java#L1717

zuul-core/src/main/java/com/netflix/zuul/netty/server/ssl/SslHandshakeInfoHandler.java
deeptiv1991: It seems like in ZuulPskServer, it is always set to false in notifyHandshakeBeginning and to true notifyHandshakeComplete. So it should always be set.

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/ZuulPskServer.java
deeptiv1991: I added code to send event SslCloseCompletionEvent on close_notify alert

zuul-core/src/main/java/com/netflix/zuul/netty/server/ssl/SslHandshakeInfoHandler.java
deeptiv1991: Actually talked to Sunny and this code is more generic and independent of ZuulPskServer. Made the change to default to false

zuul-core/src/main/java/com/netflix/zuul/netty/server/psk/ZuulPskServer.java
deeptiv1991: Sunny asked the BC folks but we havent received the response yet
https://github.com/bcgit/bc-java/issues/1673

deeptiv1991: Todo Deepti : Why we are using @SneakyThrows

zuul-core/src/main/java/com/netflix/netty/common/ssl/SslHandshakeInfo.java
```diff
@@ -31,20 +32,26 @@ public class SslHandshakeInfo {
...
     public SslHandshakeInfo(
             boolean isOfIntermediary,
             String protocol,
             String cipherSuite,
             ClientAuth clientAuthRequirement,
             Certificate serverCertificate,
-            X509Certificate clientCertificate) {
+            X509Certificate clientCertificate,
+            boolean usingExternalPSK,
+            ClientPSKIdentityInfo clientPSKIdentityInfo) {
```
jguerra, line 46: consider keeping the original constructor for backwards compatibility, and adding a new constructor with these two new parameters

jguerra (approved):

PR #1778, opened 2024-05-23
Fix fetching body content after buffer has been read

gavinbunney:
When fetching the body content of a `ZuulMessage`, the returned content differs based on if the body content chunks have been read, i.e. after it has been proxied to an upstream origin.

This change uses the `writerIndex` to instead track the total number of readable bytes of the bytebuf, so the entire content is always returned by the `getBody*` methods.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
@@ -147,16 +147,13 @@ public class ZuulMessageImpl implements ZuulMessage {
             return null;
         }
 
-        int size = 0;
-        for (final HttpContent chunk : bodyChunks) {
-            size += chunk.content().readableBytes();
-        }
+        int size = this.getBodyLength();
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/ZuulMessageImpl.java
```diff
@@ -147,16 +147,13 @@ public byte[] getBody() {
...
-            size += chunk.content().readableBytes();
-        }
+        int size = this.getBodyLength();
         final byte[] body = new byte[size];
         int offset = 0;
         for (final HttpContent chunk : bodyChunks) {
             final ByteBuf content = chunk.content();
-            final int len = content.readableBytes();
-            content.getBytes(content.readerIndex(), body, offset, len);
+            final int len = content.writerIndex(); // writer idx tracks the total readable bytes in the buffer
```
argha-c, line 155: If we do this, let's at least indicate in the comments that the API shall always return the complete buffered body, and not the pending readable bytes. Same for `getBodyLength()`

PR #1788, opened 2024-07-16
Extract methods for easier reuse/overriding

jguerra:


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/PerServerConnectionPool.java
```diff
@@ -281,6 +273,27 @@ protected void tryMakingNewConnection(
         }
     }
 
+    protected boolean checkConnectionLimit(Promise<PooledConnection> promise) {
```
gavinbunney, line 276: Maybe `s/checkConnectionLimit/withinConnectionLimit` ? Reads a bit nicer at the callsite

```
if (!withinConnectionLimit(promise)) {
...
}
```

or event `isWithinConnectionLimit` to mimic the waterline func below

gavinbunney (approved):

PR #1803, opened 2024-08-19
Make sure promise finishes during graceful shutdowns

jguerra:
The graceful close promise will not finish if some channels error out while closing. Make sure to finish the promise even if some channels have errors during closing

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientConnectionsShutdown.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientConnectionsShutdown.java
@@ -30,10 +30,9 @@ import io.netty.channel.group.ChannelGroupFuture;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientConnectionsShutdown.java
```diff
@@ -113,7 +113,14 @@ Promise<Void> gracefullyShutdownClientChannels(boolean forceCloseAfterTimeout) {
             ScheduledFuture<?> timeoutTask = executor.schedule(
                     () -> {
                         LOG.warn("Force closing remaining {} active client channels.", channels.size());
-                        channels.close();
+                        channels.close().addListener(future -> {
```
argha-c, line 116: This shouldn't be necessary.

>The graceful close promise will not finish if some channels error out while closing. 

The `closeFuture` should be invoked even if certain channels error out on close(). It has the listener wired to set a success on the Promise. Where do you see this not being set?

argha-c: That's really odd. Anyway, this doesn't hurt 🚢

argha-c (approved):

PR #1804, opened 2024-08-20
Update latest deps

argha-c:
Looks like we haven't updated these in a while.


jguerra:
The sample module is breaking - maybe discovery removed guice support?

PR #1816, opened 2024-09-09
Revert "build(deps): bump com.netflix.eureka:eureka-client from 1.10.18 to 2.0.3"

lalernehl:
Reverts Netflix/zuul#1791


argha-c:
Context: unblock the Netty upgrade while we work around the Eureka client guice support deprecation.

PR #1824, opened 2024-09-16
Add new retry policy adjustment method

jguerra:
Adds a more specific adjustRetryPolicyIfNeeded for cases when we have a HttpResponse from the origin (e.g. origin returned 503)

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/origins/NettyOrigin.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/origins/NettyOrigin.java
@@ -26,7 +26,10 @@ import com.netflix.zuul.netty.connectionpool.PooledConnection;
 import com.netflix.zuul.niws.RequestAttempt;
 import com.netflix.zuul.passport.CurrentPassport;
 import io.netty.channel.EventLoop;
+import io.netty.handler.codec.http.HttpResponse;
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/origins/NettyOrigin.java
```diff
@@ -77,4 +80,8 @@ RequestAttempt newRequestAttempt(
     IClientConfig getClientConfig();
 
     Registry getSpectatorRegistry();
+
+    default void adjustRetryPolicyIfNeeded(HttpRequestMessage zuulRequest, @Nullable HttpResponse response) {
+        adjustRetryPolicyIfNeeded(zuulRequest);
```
argha-c, line 85: Can we avoid this overload by using an `||` or the original method and a new one derived from the response?
That feels cleaner imo.

jguerra: I added a new (awkwardly named) method `originRetryPolicyAdjustmentIfNeeded`

argha-c: Naming is hard. It just needs to convey that the adjustment is inferred by the origin response.

argha-c (approved):

PR #1843, opened 2024-11-03
Add errorprone and nullaway analysis

connorworley:
This currently spews a bunch of warnings, which we can reduce over time. At a minimum, new changes should avoid introducing new warnings.


zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
```diff
@@ -157,11 +157,6 @@ public SocketAddress apply(String input) {
             }
             throw new AssertionError("Missed cased: " + bindType);
         }
-
-        @Override
-        public boolean equals(Object object) {
-            return false;
```
jguerra, line 163: 😅

jguerra (approved):
Sweet

zuul-core/src/main/java/com/netflix/zuul/netty/server/SocketAddressProperty.java
connorworley: Yeah not sure why this was here, errorprone complains that it doesn't implement hashCode as well but I'm not sure we need it in the first place.

PR #1844, opened 2024-11-04
Add Key<T> specialization to SessionContext.containsKey

connorworley:


zuul-core/src/test/java/com/netflix/zuul/context/SessionContextTest.java
```diff
@@ -107,4 +105,18 @@ void remove() {
         Truth.assertThat(context.get(key)).isNull();
         Truth.assertThat(val).isEqualTo("bar");
     }
+
+    @Test
+    void containsKey() {
```
jguerra, line 110: maybe add a test case for the non-typed version as well (if it doesn't already exist)?

jguerra (approved):

PR #1852, opened 2024-12-09
Allow ACCESS logs fields overrides

lalernehl:
Add `AccessLogInboundChannelHandler ` field overrides to allow child classes to inject data.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/netty/common/accesslog/AccessLogChannelHandler.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/accesslog/AccessLogChannelHandler.java
@@ -39,6 +39,8 @@ public final class AccessLogChannelHandler {
     private static final AttributeKey<RequestState> ATTR_REQ_STATE =
             AttributeKey.newInstance("_accesslog_requeststate");
 
+    private static final AttributeKey<String> ATTR_CLIENT_IP = AttributeKey.valueOf("client_ip");
+
     private static final Logger LOG = LoggerFactory.getLogger(AccessLogChannelHandler.class);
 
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/netty/common/accesslog/AccessLogPublisher.java
```diff
@@ -59,13 +59,15 @@ public void log(
...
             Integer localPort,
             String remoteIp,
+            String originalRemoteIp,
             Long durationNs,
             Integer requestBodySize,
             Integer responseBodySize) {
         StringBuilder sb = new StringBuilder();
 
         String dateTimeStr = dateTime != null ? dateTime.format(DATE_TIME_FORMATTER) : "-----T-:-:-";
         String remoteIpStr = (remoteIp != null && !remoteIp.isEmpty()) ? remoteIp : "-";
```
lalernehl, line 69: Decided to keep this one as it can provide extra visibility. For example when requests travel through CLBs

zuul-core/src/main/java/com/netflix/netty/common/accesslog/AccessLogChannelHandler.java
```diff
@@ -100,6 +109,7 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                         state.dateTime,
                         localPort,
                         remoteIp,
```
lalernehl, line 111: Decided to keep this one as it can provide extra visibility. For example when requests travel through CLBs

jguerra (approved):

PR #1857, opened 2024-12-16
Maintain order of query params

jguerra:
Switch from `ArrayListMultimap` to `LinkedListMultimap`. Both of these data structures are `ListMultimap`, but  `LinkedListMultimap` supports deterministic iteration order. Query parameter order is now preserved between parsing and rebuilding

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
@@ -16,13 +16,13 @@
 package com.netflix.zuul.message.http;
 
 import com.google.common.base.Strings;
-import com.google.common.collect.ArrayListMultimap;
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/message/http/HttpQueryParams.java
```diff
@@ -39,10 +39,10 @@
 public class HttpQueryParams implements Cloneable {
     private final ListMultimap<String, String> delegate;
     private final boolean immutable;
-    private final HashMap<String, Boolean> trailingEquals;
+    private final Map<String, Boolean> trailingEquals;
 
     public HttpQueryParams() {
-        delegate = ArrayListMultimap.create();
+        delegate = LinkedListMultimap.create();
```
lalernehl, line 45: Very nice!

lalernehl (approved):
LGTM

PR #1877, opened 2025-01-16
Fix broken log statements

jguerra:
I ran the inspection over the whole project, and there were no more instances of this


lalernehl (approved):
LGTM

PR #1881, opened 2025-01-27
Fix GitHub action environment variable substitution 

lociko:
GitHub Action should not be vulnerable to RCE via unsanitized `version` input. 
For example, malicious actors (contributor) can execute the workflow with `$(printenv|base64)` which will expose secrets from ENVs.


jguerra (approved):
Thanks for the contribution

PR #1885, opened 2025-02-05
Make 503s ribbon request success

lalernehl:
No longer treat 503s as a special case.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -23,6 +23,7 @@ import com.google.errorprone.annotations.ForOverride;
 import com.netflix.client.ClientException;
 import com.netflix.client.config.IClientConfigKey;
 import com.netflix.config.CachedDynamicLongProperty;
+import com.netflix.config.DynamicBooleanProperty;
 import com.netflix.config.DynamicIntegerSetProperty;
 import com.netflix.netty.common.ByteBufUtil;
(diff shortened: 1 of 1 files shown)


jguerra (approved):

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -151,6 +152,8 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
     public static final Set<String> IDEMPOTENT_HTTP_METHODS = Sets.newHashSet("GET", "HEAD", "OPTIONS");
     private static final DynamicIntegerSetProperty RETRIABLE_STATUSES_FOR_IDEMPOTENT_METHODS =
             new DynamicIntegerSetProperty("zuul.retry.allowed.statuses.idempotent", "500");
+    private static final DynamicBooleanProperty ENABLE_ORIGIN_THROTTLED_FOR_LB_ERRORS =
+            new DynamicBooleanProperty("zuul.lb.503error.enabled", false);
```
gavinbunney, line 156: Maybe something like:

```suggestion
zuul.proxy.origin.errors.include.throttled
```

So it can be reused by (other) ribbon changes

lalernehl: I think having `lb/load balancer` in the name somehow, gives more context. In either case, I don't think we need this property once I roll this out, my plan is to clean it up after.

gavinbunney (approved):

PR #1894, opened 2025-02-17
Upgrade Netty version to 4.1.118

spethani:
Details about version 4.1.118: https://netty.io/news/2025/02/10/4-1-118-Final.html


spethani:
Looking into build failure; seems to be an issue with the `junit` dependencies.

PR #1903, opened 2025-02-28
Move connection pool metric creation to a separate class for easier reuse

jguerra:
DefaultClientChannelManager creates a ton of counters that get passed around to various connection pooling classes. I created a new data class that can hold all of the various counters. I didn't try and tackle the ugly methods that take all of the counters as an argument since i'd have to make a lot of breaking changes to change them

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ConnectionPoolMetrics.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ConnectionPoolMetrics.java
@@ -46,30 +46,28 @@ public record ConnectionPoolMetrics(Counter createNewConnCounter,
                                     AtomicInteger connsInPool,
                                     AtomicInteger connsInUse) {
 
-    public static final String METRIC_PREFIX = "connectionpool_";
-
     public static ConnectionPoolMetrics create(OriginName originName, Registry registry) {
-        Counter createNewConnCounter = newCounter("create", originName, registry);
-        Counter createConnSucceededCounter = newCounter("create_success", originName, registry);
-        Counter createConnFailedCounter = newCounter("create_fail", originName, registry);
+        Counter createNewConnCounter = newCounter("connectionpool_create", originName, registry);
+        Counter createConnSucceededCounter = newCounter("connectionpool_create_success", originName, registry);
+        Counter createConnFailedCounter = newCounter("connectionpool_create_fail", originName, registry);
 
-        Counter closeConnCounter = newCounter("close", originName, registry);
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ConnectionPoolMetrics.java
```diff
@@ -0,0 +1,91 @@
...
+        AtomicInteger connsInUse = newGauge("inUse", originName, registry);
+
+        return new ConnectionPoolMetrics(createNewConnCounter, createConnSucceededCounter, createConnFailedCounter,
+                closeConnCounter, closeAbovePoolHighWaterMarkCounter, closeExpiredConnLifetimeCounter, requestConnCounter,
+                reuseConnCounter, releaseConnCounter, alreadyClosedCounter, connTakenFromPoolIsNotOpen,
+                maxConnsPerHostExceededCounter, closeWrtBusyConnCounter, circuitBreakerClose, connEstablishTimer, connsInPool, connsInUse);
+    }
+
+    private static Counter newCounter(String metricName, OriginName originName, Registry registry) {
+        return registry.counter(METRIC_PREFIX + metricName, "id", originName.getMetricId());
```
gavinbunney, line 81: Could you perhaps inline this into the `create` method instead, and remove the `METRIC_PREFIX +` concat? Normally when I'm searching around for where a metric is emitted, it's much easier to just match on the whole name, e.g. `connectionpool_create`, so if it's just all listed there it is a lot easier to locate

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/DefaultClientChannelManager.java
```diff
@@ -61,48 +60,22 @@
...
-    private final Counter closeWrtBusyConnCounter;
-    private final Counter circuitBreakerClose;
-
-    private final PercentileTimer connEstablishTimer;
-    private final AtomicInteger connsInPool;
-    private final AtomicInteger connsInUse;
-
     private final ConcurrentHashMap<DiscoveryResult, IConnectionPool> perServerPools;
-
+    private final ConnectionPoolMetrics metrics;
```
gavinbunney, line 74: ❤️

zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ConnectionPoolMetrics.java
jguerra: for sure, I find the prefixes annoying as well

gavinbunney (approved):

PR #1928, opened 2025-04-03
Add timing metric for client request headers

connorworley:
Turns out we don't have a timing metric around header reading internally, so this change adds one.


zuul-core/src/main/java/com/netflix/zuul/netty/timeouts/HttpHeadersTimeoutHandler.java
```diff
@@ -84,12 +90,13 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
             public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
                 try {
                     if (msg instanceof HttpMessage) {
+                        httpHeadersReadTimer.record(System.nanoTime() - ctx.channel().attr(HTTP_HEADERS_READ_START_TIME).get(), TimeUnit.NANOSECONDS);
```
lindseyreynolds, line 93: Where is this information emitted? Is there a metric that it is tied to?

lindseyreynolds (approved):

PR #1929, opened 2025-04-07
Extract client ip method

lindseyreynolds:
There are cases where the client ip set in the http request object is not correct. We should use the ATTR_CLIENT_IP value  set in ClientIpChannelHandler instead, which takes into account ppv2 and XFF headers. 
The only issue is that this value is scoped to edgezuul-netty, but the http request is created in zuul oss in ClientRequestReceiver, which means it is not accessible. 
To deal with this, we can extract the code to get the client ip and then override it in edgezuul-netty by creating an EdgeClientRequestReceiver class that returns the client ip using ClientIpChannelHandler.ATTR_CLIENT_IP instead. That PR will follow this one.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
@@ -321,7 +321,7 @@ public class ClientRequestReceiver extends ChannelDuplexHandler {
             context = new SessionContext();
         }
 
-        // Get the client IP
+        // Get the client IP (ignore XFF headers at this point, as that can be app specific).
         Channel channel = clientCtx.channel();
         String clientIp = getClientIp(channel);
 
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
@@ -16,9 +16,6 @@
 
 package com.netflix.zuul.netty.server;
 
-import static com.netflix.netty.common.HttpLifecycleChannelHandler.CompleteEvent;
-import static com.netflix.netty.common.HttpLifecycleChannelHandler.CompleteReason;
-
 import com.netflix.netty.common.SourceAddressChannelHandler;
 import com.netflix.netty.common.ssl.SslHandshakeInfo;
 import com.netflix.netty.common.throttle.RejectionUtils;
@@ -66,6 +63,10 @@ import io.netty.util.AttributeKey;
 import io.netty.util.ReferenceCountUtil;
 import io.perfmark.PerfMark;
 import io.perfmark.TaskCloseable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.net.ssl.SSLException;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


lalernehl (approved):
NIT: Add a small test to verify the clientIP comes from `ATTR_SOURCE_ADDRESS` and in the child class we can verify then that it comes from `ATTR_CLIENT_IP` instead?

lindseyreynolds:
> NIT: Add a small test to verify the clientIP comes from `ATTR_SOURCE_ADDRESS` and in the child class we can verify then that it comes from `ATTR_CLIENT_IP` instead?

This PR doesnt actually change anything, but I will add tests to the new `EdgeClientIpChannelHandlerTest` class, where I verify that the client ip is being populated by the new attribute.

lindseyreynolds:
> NIT: Add a small test to verify the clientIP comes from `ATTR_SOURCE_ADDRESS` and in the child class we can verify then that it comes from `ATTR_CLIENT_IP` instead?

lol I thought about that and then saw that no test classes exist, so didnt feel like this refactor was worth the time to set things up.
I was thinking that just adding tests to the new `EdgeClientIpChannelHandlerTest` class, where I verify that the client ip is being populated by the new attribute will suffice. What do you think?

lalernehl:
> > NIT: Add a small test to verify the clientIP comes from `ATTR_SOURCE_ADDRESS` and in the child class we can verify then that it comes from `ATTR_CLIENT_IP` instead?
> 
> lol I thought about that and then saw that no test classes exist, so didnt feel like this refactor was worth the time to set things up. I was thinking that just adding tests to the new `EdgeClientIpChannelHandlerTest` class, where I verify that the client ip is being populated by the new attribute will suffice. What do you think?

I think this is the test class, isn't it? https://github.com/Netflix/zuul/blob/master/zuul-core/src/test/java/com/netflix/zuul/netty/server/ClientRequestReceiverTest.java

lindseyreynolds:
> > > NIT: Add a small test to verify the clientIP comes from `ATTR_SOURCE_ADDRESS` and in the child class we can verify then that it comes from `ATTR_CLIENT_IP` instead?
> > 
> > 
> > lol I thought about that and then saw that no test classes exist, so didnt feel like this refactor was worth the time to set things up. I was thinking that just adding tests to the new `EdgeClientIpChannelHandlerTest` class, where I verify that the client ip is being populated by the new attribute will suffice. What do you think?
> 
> I think this is the test class, isn't it? https://github.com/Netflix/zuul/blob/master/zuul-core/src/test/java/com/netflix/zuul/netty/server/ClientRequestReceiverTest.java

What the heck - I think IntelliJ was having issues because this class wasnt showing up when I searched through the files earlier. Thank you for calling this out!

lalernehl:
> > > > NIT: Add a small test to verify the clientIP comes from `ATTR_SOURCE_ADDRESS` and in the child class we can verify then that it comes from `ATTR_CLIENT_IP` instead?
> > > 
> > > 
> > > lol I thought about that and then saw that no test classes exist, so didnt feel like this refactor was worth the time to set things up. I was thinking that just adding tests to the new `EdgeClientIpChannelHandlerTest` class, where I verify that the client ip is being populated by the new attribute will suffice. What do you think?
> > 
> > 
> > I think this is the test class, isn't it? https://github.com/Netflix/zuul/blob/master/zuul-core/src/test/java/com/netflix/zuul/netty/server/ClientRequestReceiverTest.java
> 
> What the heck - I think IntelliJ was having issues because this class wasnt showing up when I searched through the files earlier. Thank you for calling this out!

I also found it confusing sometimes the way the path is shown in intelliJ 🤷

PR #1947, opened 2025-05-20
Resume filter chain in onComplete instead of onNext

jguerra:
Starting the next filter in onNext prevents onComplete from running until the next async filter is encountered

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
@@ -55,7 +55,8 @@ import rx.functions.Action1;
 import rx.schedulers.Schedulers;
 
 /**
- * Subclasses of this class are supposed to be thread safe and hence should not have any non final member variables
+ * Subclasses of this class are supposed to be thread safe
+ *
  * Created by saroskar on 5/18/17.
  */
 @ThreadSafe
@@ -202,7 +203,6 @@ public abstract class BaseZuulFilterRunner<I extends ZuulMessage, O extends Zuul
                 filterRunStatus = ExecutionStatus.SKIPPED;
             }
 
-            ;
             try (TaskCloseable ignored2 = PerfMark.traceTask(filter, f -> f.filterName() + ".shouldSkipFilter")) {
                 if (shouldSkipFilter(inMesg, filter)) {
                     filterRunStatus = ExecutionStatus.SKIPPED;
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -239,7 +239,7 @@ protected final O filter(ZuulFilter<I, O> filter, I inMesg) {
             inMesg.runBufferedBodyContentThroughFilter(filter);
 
             if (filter.getSyncType() == FilterSyncType.SYNC) {
-                SyncZuulFilter<I, O> syncFilter = (SyncZuulFilter) filter;
+                SyncZuulFilter<I, O> syncFilter = (SyncZuulFilter<I, O>) filter;
```
argha-c, line 242: 👍🏼

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -474,7 +475,12 @@ public void onError(Throwable ex) {
         public void onCompleted() {
             try (TaskCloseable ignored = PerfMark.traceTask(filter, f -> f.filterName() + ".onCompletedAsync")) {
                 PerfMark.linkIn(onCompletedLinkOut.get());
+                if (outMesg == null) {
+                    outMesg = filter.getDefaultOutput(inMesg);
+                }
                 decrementConcurrency();
+                recordFilterCompletion(ExecutionStatus.SUCCESS, filter, startTime, inMesg, snapshot);
+                resumeInBindingContext(outMesg, filter.filterName());
```
argha-c, line 483: A nice side-effect of this may be `onCompleted()` actually guarantees exactly once semantics.

argha-c (approved):
Really good find! Excited to see if it helps with other scheduling bottlenecks.

zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -474,7 +475,12 @@ public void onError(Throwable ex) {
         public void onCompleted() {
             try (TaskCloseable ignored = PerfMark.traceTask(filter, f -> f.filterName() + ".onCompletedAsync")) {
                 PerfMark.linkIn(onCompletedLinkOut.get());
+                if (outMesg == null) {
```
argha-c, line 478: And I think this is a non-issue, but does resuming `onCompleted` affect the mutation order of `outMesg`?

jguerra: I don't think there should be any issues. After reviewing things again I did end up adding error handling to onComplete, as the request would hang if an exception was thrown for some reason

PR #1948, opened 2025-05-20
Custom scheduler for async filters

jguerra:


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -365,8 +367,15 @@ public HttpResponseMessage apply(HttpRequestMessage input) {
     @Override
     public HttpContent processContentChunk(ZuulMessage zuulReq, HttpContent chunk) {
         if (originConn != null) {
+            if (chunk instanceof LastHttpContent && !receivedChunkAfterProxyStarted) {
```
jguerra, line 370: this is a subtly change. If a last content check comes in after a connection has been established (e.g. a pooled connection is used), and no other content chunks have arrived in this state, then buffer the last http content so requests can be eligible for retries

zuul-integration-test/src/test/java/com/netflix/zuul/integration/IntegrationTest.java
```diff
@@ -117,6 +128,8 @@ static void beforeAll() {
         int wireMockPort = wireMockExtension.getPort();
         AbstractConfiguration config = ConfigurationManager.getConfigInstance();
         config.setProperty("zuul.server.netty.socket.force_nio", "true");
+        // run with a single event loop to make tests more deterministic, especially for connection pooling
+        config.setProperty("zuul.server.netty.threads.worker", "1");
```
gavinbunney, line 132: Should we also force at-least one test to run with 1 vs 2 workers?

gavinbunney, line 132: Should we also force at-least one test to run with 1 vs 2 workers?

gavinbunney (approved):

zuul-core/src/test/java/com/netflix/zuul/netty/filter/EventExecutorSchedulerTest.java
```diff
@@ -0,0 +1,114 @@
...
+            Action0 action = () -> {
+                executed.set(true);
+            };
+            Subscription schedule = worker.schedule(action);
+            assertTrue(executed.get());
+            assertTrue(schedule.isUnsubscribed());
+            latch.countDown();
+        });
+
+        if (!Uninterruptibles.awaitUninterruptibly(latch, 5, TimeUnit.SECONDS)) {
```
lalernehl, line 80: Why not `latch.await`? To avoid the `InterruptedException`?

lalernehl (approved):

zuul-core/src/test/java/com/netflix/zuul/netty/filter/EventExecutorSchedulerTest.java
jguerra: yeah, just a habit of mine at this point.

zuul-integration-test/src/test/java/com/netflix/zuul/integration/IntegrationTest.java
jguerra: yeah, that's a good call

PR #1962, opened 2025-06-17
Inside PushMessageSender, add method for logging body

AlexanderEllis:
This is just another method that a child class can overwrite to do additional logging with the request and body.


AlexanderEllis (commented):
Looks great!

PR #1967, opened 2025-07-07
Fix header size exceeded error

lalernehl:


zuul-core/src/main/java/com/netflix/zuul/netty/NettyRequestAttemptFactory.java
```diff
@@ -62,6 +63,10 @@ public ErrorType mapNettyToOutboundErrorType(Throwable t) {
             return OutboundErrorType.RESET_CONNECTION;
         }
 
+        if (t instanceof HeaderListSizeException) {
```
jguerra, line 66: Can you add a test for this?

jguerra (approved):

zuul-core/src/main/java/com/netflix/zuul/netty/NettyRequestAttemptFactory.java
lalernehl: added

PR #2018, opened 2025-10-22
Fix query parsing in massageRequestURI

lindseyreynolds:
Currently, if query param values are encoded, this method decodes them and pulls out any hidden query params before sending them off to the origin. 

Example:
- A request arrives with this query param: `param=123%26hidden%3Dvalue`
- ProxyEndpoint decodes this query param and converts it to `param=123` and `hidden=value`
- These two query params get sent to the origin

This PR updates this method to just decode the values, but not separate them out. 

I confirmed that the old code passes all the tests I added (aside from the ones with ampersands) to ensure that no other behavior is being impacted by this change.

Jira: https://netflix.atlassian.net/browse/CGW-2721

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -157,7 +157,7 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
     private static final DynamicIntegerSetProperty RETRIABLE_STATUSES_FOR_IDEMPOTENT_METHODS =
             new DynamicIntegerSetProperty("zuul.retry.allowed.statuses.idempotent", "500");
     private static final DynamicBooleanProperty ENABLE_MODERN_QUERY_PARSING =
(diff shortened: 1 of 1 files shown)


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -154,6 +156,8 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
     public static final Set<String> IDEMPOTENT_HTTP_METHODS = Sets.newHashSet("GET", "HEAD", "OPTIONS");
     private static final DynamicIntegerSetProperty RETRIABLE_STATUSES_FOR_IDEMPOTENT_METHODS =
             new DynamicIntegerSetProperty("zuul.retry.allowed.statuses.idempotent", "500");
+    private static final DynamicBooleanProperty ENABLE_MODERN_QUERY_PARSING =
+            new DynamicBooleanProperty("zuul.enable.modern.query.parsing", true);
```
lalernehl: nit: should the format of this be more like `zuul.feature.enabled`. I think many of our existing properties follow this pattern.

PR #2038, opened 2025-12-11
Add flag for matching on excessive filter execution time

jguerra:
For easy matching with mantis


zuul-core/src/main/java/com/netflix/zuul/netty/filter/BaseZuulFilterRunner.java
```diff
@@ -349,6 +349,7 @@ protected void recordFilterCompletion(
         long execTimeNs = System.nanoTime() - startTime;
         long execTimeMs = execTimeNs / 1_000_000L;
         if (execTimeMs >= FILTER_EXCESSIVE_EXEC_TIME.get()) {
+            zuulCtx.setEventProperty("excessive_filter_execution_time", true);
```
jguerra:
```suggestion
            zuulCtx.setEventProperty("filter_execution_time_exceeded", true);
```

PR #2072, opened 2026-02-13
Introduce the idea of filter constraints 

jguerra:
Adds a new annotation based mechanism for adding restrictions around when a filter should run. The purpose is to enable centralizing common shouldFilter logic


gavinbunney (approved):

zuul-core/src/main/java/com/netflix/zuul/filters/ZuulFilter.java
```diff
@@ -76,6 +77,11 @@ default FilterCategory category() {
         }
     }
 
+    default Class<? extends FilterConstraint>[] constraints() {
+        Filter annotation = getClass().getAnnotation(Filter.class);
```
gavinbunney: This could NPE if the filter is missing the constraint annotation; not sure if we should throw a formal exception or just return an empty array

gavinbunney (approved):

PR #2090, opened 2026-02-25
Prefer Objects.requireNonNull or @NonNull

jguerra:
`com.netflix.spectator.impl.Preconditions` isn't recognized by intellij, and resulted in null warnings. I replaced that version of Preconditions and Guava's with either lombok or the Objects equivalent


gavinbunney (approved):
👏

PR #2127, opened 2026-04-29
Return bad request for invalid URI

fool1280:
- Decode %2E before normalization: per RFC 3986 §2.4, percent-encoded unreserved characters (like .) should be decoded before processing, so dot-segments like %2E%2E are properly collapsed by normalize()    
- Propagate URISyntaxException from parsePath: instead of silently swallowing parse errors and falling back to a regex, invalid URIs now set a BAD_URI flag on the context and are rejected with a 400         
- Remove the fallback regex: deleted URL_REGEX and the manual path-parsing fallback; invalid URIs are now a hard reject rather than a best-effort parse     
- Remove opaque URI handling

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
@@ -94,7 +94,6 @@ public class ClientRequestReceiver extends ChannelDuplexHandler {
     private static final Logger LOG = LoggerFactory.getLogger(ClientRequestReceiver.class);
     private static final String SCHEME_HTTP = "http";
     private static final String SCHEME_HTTPS = "https";
-    private static final String BAD_URI = "bad_uri";
 
     private final SessionContextDecorator decorator;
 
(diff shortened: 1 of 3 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientRequestReceiver.java
```diff
@@ -95,10 +94,7 @@ public class ClientRequestReceiver extends ChannelDuplexHandler {
     private static final Logger LOG = LoggerFactory.getLogger(ClientRequestReceiver.class);
     private static final String SCHEME_HTTP = "http";
     private static final String SCHEME_HTTPS = "https";
-
-    // via @stephenhay https://mathiasbynens.be/demo/url-regex, groups added
-    // group 1: scheme, group 2: domain, group 3: path+query
-    private static final Pattern URL_REGEX = Pattern.compile("^(https?)://([^\\s/$.?#].[^\\s/]*)([^\\s]*)$");
+    private static final String BAD_URI = "bad_uri";
```
jguerra: can you switch this to the type safe variety (SessionContext.Key)

jguerra (approved):

PR #2129, opened 2026-05-05
Correctly handle http conn state when origin responds before `LastHttpContent` is sent

gavinbunney:
This PR ensures we close the origin connection (rather than returning it to the pool) when an HTTP/1.1 origin sends a response before zuul finishes streaming the request body to it.

If an origin responds before zuul finishes sending the request body without signally to close the connection, the outbound `HttpClientCodec` encoder is left mid-body (state `ST_CONTENT_NON_CHUNK`). When the connection is reused for the next request, the encoder rejects it with `IllegalStateException: unexpected message type: DefaultHttpRequest, state: 1`, resulting in a HTTP 500 response to an unrelated request.

Per RFC 9112 / RFC 7230, an origin that produces a response without consuming the full request body is expected to signal `Connection: close` so the connection is not reused. However, some origins don't, so we ensure the connection pool correctly protects against this case.

### DIFF
--- a/zuul-core/src/test/java/com/netflix/netty/common/HttpClientLifecycleChannelHandlerTest.java
+++ b/zuul-core/src/test/java/com/netflix/netty/common/HttpClientLifecycleChannelHandlerTest.java
@@ -25,14 +25,26 @@ import io.netty.handler.codec.http.DefaultHttpRequest;
 import io.netty.handler.codec.http.DefaultLastHttpContent;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpVersion;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 class HttpClientLifecycleChannelHandlerTest {
 
+    private EmbeddedChannel channel;
+
+    @BeforeEach
+    void setup() {
+        channel = new EmbeddedChannel(HttpClientLifecycleChannelHandler.OUTBOUND_CHANNEL_HANDLER);
+    }
+
+    @AfterEach
+    void tearDown() {
+        channel.finishAndReleaseAll();
+    }
+
     @Test
     void lastContentPendingAfterRequestHeadersOnly() {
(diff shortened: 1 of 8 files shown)


zuul-integration-test/src/test/java/com/netflix/zuul/integration/EarlyOriginResponseConnectionReuseTest.java
```diff
@@ -0,0 +1,223 @@
...
+        // consume the rest, leaving the outbound HttpClientCodec encoder mid-body
+        String partialPost = "POST /redirect-on-headers HTTP/1.1\r\n"
+                + "Host: localhost:" + zuulPort + "\r\n"
+                + "Content-Type: application/octet-stream\r\n"
+                + "Content-Length: 2000\r\n"
+                + "Connection: keep-alive\r\n"
+                + "\r\n"
+                + "X".repeat(100);
+
+        try (Socket clientSocket = new Socket("localhost", zuulPort)) {
```
jguerra, line 137: 👍

zuul-core/src/test/java/com/netflix/netty/common/HttpClientLifecycleChannelHandlerTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.jupiter.api.Test;
+
+class HttpClientLifecycleChannelHandlerTest {
+
+    @Test
+    void lastContentPendingAfterRequestHeadersOnly() {
+        EmbeddedChannel channel = new EmbeddedChannel(HttpClientLifecycleChannelHandler.OUTBOUND_CHANNEL_HANDLER);
+
+        channel.writeOutbound(new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.POST, "/foo"));
+        channel.writeOutbound(new DefaultHttpContent(Unpooled.wrappedBuffer(new byte[] {1, 2, 3})));
```
jguerra, line 49: Can you release these? I _think_ we have some tests that check for leaks in zuul-core

jguerra (approved):

PR #2131, opened 2026-05-12
Preserve failure nf status set by recordFinalError in buildZuulHttpResponse

lindseyreynolds:
When origin.recordFinalError() sets a specific failure StatusCategory (e.g. from an origin subclass override), the subsequent unconditional setStatusCategory call was overwriting it with the generic status derived from the response code. 

This PR switches to use storeStatusCategoryIfNotAlreadyFailure so that a failure status set by recordFinalError is not overwritten.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -910,7 +910,7 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
         // Collect some info about the received response.
         origin.recordFinalResponse(zuulResponse);
         origin.recordFinalError(zuulRequest, ex);
-        StatusCategoryUtils.setStatusCategory(zuulCtx, statusCategory);
+        StatusCategoryUtils.storeStatusCategoryIfNotAlreadyFailure(zuulCtx, statusCategory);
         zuulCtx.setError(ex);
         zuulCtx.put("origin_http_status", Integer.toString(respStatus));
 


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -910,7 +910,7 @@ private HttpResponseMessage buildZuulHttpResponse(
         // Collect some info about the received response.
         origin.recordFinalResponse(zuulResponse);
         origin.recordFinalError(zuulRequest, ex);
-        StatusCategoryUtils.setStatusCategory(zuulCtx, statusCategory);
+        StatusCategoryUtils.storeStatusCategoryIfNotAlreadyFailure(zuulCtx, statusCategory);
```
gavinbunney, line 913: I wonder if this handles retries correctly? Since I think this block is called twice for a retry scenario, first time through sets it to error X, second time through the error might be Y now?

lindseyreynolds: Yep it does. I was able to test this by running wall-e, but not running the sidecar. Then I sent a curl and put a break point on this line and in postErrorProcessing to confirm the failure. Then I started the sidecar while it was paused and continued the request. When it got to this line, statusCategory had been updated to SUCCESS, since that was the actual status of the current request. It did not hold onto the previous retry value.

lindseyreynolds: I also have an integration test that shows this in my timeout branch, but cant push it until this oss change merges because I am using the oss directory with this change in place.

gavinbunney (approved):

PR #2137, opened 2026-05-22
Fix for connection leak issue with HTTP 1xx responses

matthoffman:
We were returning the origin connection to the pool after getting a 1xx HTTP response, which was incorrect – those are responses that indicate a further response was coming. In some cases, the origin's eventual (real) response could actually go to the wrong client in these cases.

This also fixes an issue where the client -> proxy connection explicitly handled HTTP 100 (Continue), but not any other 1xx response. The others aren't common, but we should handle them anyway.

IMPORTANT NOTE: with this change, we're still not passing 1xx HTTP responses from origin back to clients – they're swallowed by Zuul. I originally started on a larger change that did pass them back, but it would require more changes, and more risk. It messes with backpressure, and probably requires injecting the interim responses directly to clients, bypassing the outbound filter chain, which seems ... potentially problematic?  Meanwhile, we have no actual use case for returning interim responses to the client right now (and we never handled them correctly before, so while it's technically a behavior change, it's not one anyone would be relying on). 

Some Claude code in here, fwiw, but I've reviewed enough that I'm to blame for mistakes :)

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
@@ -860,18 +860,12 @@ public class ProxyEndpoint extends SyncZuulFilterAdapter<HttpRequestMessage, Htt
 
     /**
      * Swallows 1xx interim responses (100 Continue, 102 Processing, 103 Early Hints) rather than relaying
-     * them to the client. We have no use case that requires propagating informational responses, and
-     * forwarding them is not free: an interim response would otherwise traverse the outbound filter chain
-     * (whose progress state is single-slot per filter type in the {@link SessionContext}, so it cannot run
-     * twice per request without clobbering the final response), pick up terminal-response headers, and need
-     * special framing on both the HTTP/1.1 and HTTP/2 client paths. Swallowing avoids all of that and mirrors
-     * how Zuul already answers the client's {@code Expect: 100-continue} itself instead of relaying it.
+     * them to the client.
      *
-     * <p>The security fix this protects is unaffected: the connection is not returned to the pool here, and
-     * the lifecycle handlers keep it open for the real terminal response (they suppress SESSION_COMPLETE on a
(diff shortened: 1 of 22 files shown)


zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
```diff
@@ -838,13 +849,35 @@ public void responseFromOrigin(HttpResponse originResponse) {
...
+     * twice per request without clobbering the final response), pick up terminal-response headers, and need
+     * special framing on both the HTTP/1.1 and HTTP/2 client paths. Swallowing avoids all of that and mirrors
+     * how Zuul already answers the client's {@code Expect: 100-continue} itself instead of relaying it.
+     *
+     * <p>The security fix this protects is unaffected: the connection is not returned to the pool here, and
+     * the lifecycle handlers keep it open for the real terminal response (they suppress SESSION_COMPLETE on a
+     * 1xx). {@code pendingInterimResponse} tells {@link #filterResponseChunk} to drop the empty
+     * {@code LastHttpContent} that Netty emits after every 1xx rather than treating it as the end of the
+     * response.
+     */
```
gavinbunney, line 869: nit: can tighten this comment more around the first line - other pieces are related to the underlying fix rather than why this code exists for the future. Something more along the lines of:

```
    /**
     * Swallows 1xx interim responses (100 Continue, 102 Processing, 103 Early Hints) rather than relaying
     * them to the client. 
     *
     * <p>Note: the connection is not returned to the pool here, and
     * the lifecycle handlers keep it open for the real terminal response (they suppress SESSION_COMPLETE on a
     * 1xx). {@code pendingInterimResponse} tells {@link #filterResponseChunk} to drop the empty
     * {@code LastHttpContent} that Netty emits after every 1xx rather than treating it as the end of the
     * response.
     */
```

gavinbunney (approved):

zuul-core/src/main/java/com/netflix/zuul/filters/endpoint/ProxyEndpoint.java
matthoffman: Sounds good, done.

PR #2138, opened 2026-05-22
Enabling header validation via fast property

ilanachalom:
Enabling header validation, configurable by FP

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/Http1FramingEnforcingHandler.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/Http1FramingEnforcingHandler.java
@@ -54,11 +54,6 @@ public final class Http1FramingEnforcingHandler extends ChannelInboundHandlerAda
             return;
         }
 
-        if (req.decoderResult().isFailure()) {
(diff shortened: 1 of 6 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/Http1FramingEnforcingHandler.java
```diff
@@ -54,6 +54,11 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
             return;
         }
 
+        if (req.decoderResult().isFailure()) {
+            rejectAndClose(ctx, req, msg, "decoder failure: " + req.decoderResult().cause());
+            return;
+        }
+
```
gavinbunney: Is there another handler we can use here? The `Http1FramingEnforcingHandler` itself is conditional and feels like this decoder check/failure should live outside the framing checks

ilanachalom: Added a new handler for this instead!

gavinbunney (approved):

PR #2145, opened 2026-06-12
Ensure `SslExceptionsHandler` swallows all client SSL exceptions, not just handshake failures

gavinbunney:
`SslExceptionsHandler` only swallows TLS handshake failures, and only when wrapped exactly one level deep (`cause.getCause() instanceof SSLHandshakeException`). Other client-induced TLS decode errors - `BAD_DECRYPT`, `WRONG_VERSION_NUMBER` and friends, typically a client speaking plaintext to the TLS port or sending a corrupt record - surface as a generic `OpenSslException` (an `SSLException`, *not* an `SSLHandshakeException`) wrapped in a `DecoderException`, so those propagate through.

The result is a full stack trace logged per bad connection: on HTTP/1.1 at ERROR via `ZuulFilterChainHandler`, and on HTTP/2 parent connections at WARN once the exception reaches the pipeline tail.

```
# currently
ERROR ZuulFilterChainHandler  zuul filter chain handler caught exception on channel: ...   # h1
WARN  DefaultChannelPipeline  An exceptionCaught() event ... reached at the tail ...        # h2 parent
io.netty.handler.codec.DecoderException: ...OpenSslException: ...OPENSSL_internal:BAD_DECRYPT

# now
debug-only; server.ssl.exception.swallowed{cause=OpenSslException} incremented
```

This PR walks the cause chain for any `SSLException` rather than checking a single level for one subtype, ensuring we catch and handle all client-side TLS errors.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/netty/common/SslExceptionsHandler.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/SslExceptionsHandler.java
@@ -20,7 +20,7 @@ import com.netflix.spectator.api.Registry;
 import io.netty.channel.ChannelHandler.Sharable;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
-import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLException;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


matthoffman (approved):
LGTM

PR #2146, opened 2026-06-13
Fix potential `CoalescingBufferQueue` desync by tracking each body reader index separately

gavinbunney:
When an origin request is retried, `ProxyEndpoint.writeBufferedBodyContent` replays the same buffered `ByteBuf` chunks to each attempt with `chunk.retain()`. A plain `retain()` shares a single reader index across _every_ attempt, so if one attempt drains the body it advances the reader index another still-in-flight attempt's netty `CoalescingBufferQueue` is pointing at.

That queue can then be left reporting `readableBytes() > 0` with nothing left to produce, and `DefaultHttp2RemoteFlowController.writeAllocatedBytes` spins emitting empty 9-byte `DATA` frames (`frame.size()` never reaches 0) into the channel outbound buffer until the heap is exhausted - netty [#11959](https://github.com/netty/netty/issues/11959). We saw exactly this signature in a production heap dump: ~134M empty frames queued on a single channel.

This switches the buffered-body replay to use `chunk.retainedDuplicate()`, giving each attempt its own reader/writer index over the same memory (shared refCnt), so one attempt draining its copy can no longer move another attempt's index.

Added a new `CoalescingBufferQueueReaderIndexDesyncTest` to verify the mechanism against the real netty classes: the shared-index drain desyncs a real `CoalescingBufferQueue`, and a frame stuck in that state spins a real `DefaultHttp2RemoteFlowController` out to 200k+ empty frames - run on a separate-thread timeout with a frame cap so the spin is observable without hanging the suite.

Raised netty/netty/issues/16946 and fix in netty/netty/pull/16947 to help prevent unbounded frames.

### DIFF
--- a/zuul-core/src/test/java/com/netflix/zuul/filters/endpoint/ProxyEndpointTest.java
+++ b/zuul-core/src/test/java/com/netflix/zuul/filters/endpoint/ProxyEndpointTest.java
@@ -51,12 +51,14 @@ import com.netflix.zuul.origins.NettyOrigin;
 import com.netflix.zuul.passport.CurrentPassport;
 import com.netflix.zuul.passport.PassportItem;
 import com.netflix.zuul.passport.PassportState;
+import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.embedded.EmbeddedChannel;
 import io.netty.channel.local.LocalAddress;
 import io.netty.handler.codec.http.DefaultHttpResponse;
 import io.netty.handler.codec.http.DefaultLastHttpContent;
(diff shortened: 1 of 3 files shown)


zuul-core/src/test/java/com/netflix/zuul/filters/endpoint/ProxyEndpointTest.java
```diff
@@ -158,6 +160,51 @@ void testRetryWillResetBodyReader() {
         assertThat(new String(request.getBody(), UTF_8)).isEqualTo("Hello There");
     }
 
+    @Test
+    void bufferedBodyReplayGivesEachOriginAttemptIndependentReaderIndex() {
```
AlexanderEllis, line 164: Good tests!

AlexanderEllis (approved):
Super subtle — great catch and fix!

PR #2154, opened 2026-06-26
Add zero-allocation iteration helpers to `Headers`

gavinbunney:
`Headers` only offers `entries()` and `keySet()` for walking headers, and both allocate sets of objects per call. Particularly if you just want to find if a header exists or remove a batch of headers, requires double-loop structures in the call-sites, with all these allocated objects being thrown away immediately.

This PR adds three helpers to allow for more efficient interaction with the headers:

* `forEach(BiConsumer<String, String>)` - walk original-case name + value, the counterpart to `forEachNormalised` for paths that re-emit headers to origins/clients and must preserve case
* `anyMatchNormalised(BiPredicate<String, String>)` - short-circuiting matching
* `removeAllNormalised(BiPredicate<String, String>)` - in-place delete, the allocation-free counterpart to `removeIf` (which builds a `HeaderName` + `Map.Entry` per entry)

### DIFF
--- a/zuul-core/src/test/java/com/netflix/zuul/message/HeadersTest.java
+++ b/zuul-core/src/test/java/com/netflix/zuul/message/HeadersTest.java
@@ -109,6 +109,23 @@ class HeadersTest {
     }
 
     @Test
+    void forEach() {
+        Headers headers = new Headers();
+        headers.add("ViA", "duct");
+        headers.add("Cookie", "this=that");
+        headers.add("Cookie", "frizzle=Frazzle");
+        Map<String, List<String>> result = new LinkedHashMap<>();
+
+        headers.forEach((k, v) ->
+                result.computeIfAbsent(k, discard -> new ArrayList<>()).add(v));
+
+        assertThat(result)
+                .containsExactly(
+                        entry("ViA", Collections.singletonList("duct")),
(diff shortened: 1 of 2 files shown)


zuul-core/src/test/java/com/netflix/zuul/message/HeadersTest.java
```diff
@@ -108,6 +108,23 @@ void getFirst_headerName_returnsDefault() {
...
+        headers.add("Cookie", "frizzle=Frazzle");
+        Map<String, List<String>> result = new LinkedHashMap<>();
+
+        headers.forEach((k, v) ->
+                result.computeIfAbsent(k, discard -> new ArrayList<>()).add(v));
+
+        assertThat(result)
+                .containsExactly(
+                        entry("ViA", Collections.singletonList("duct")),
+                        entry("Cookie", Arrays.asList("this=that", "frizzle=Frazzle")));
```
jguerra, line 125: the cool kids use List.of these days

jguerra (approved):

zuul-core/src/test/java/com/netflix/zuul/message/HeadersTest.java
gavinbunney: Shame I'm not cool 😭

PR #2155, opened 2026-06-26
Build origin requests and client responses with the allocation-free `Headers` iterators

gavinbunney:
`OriginResponseReceiver` rebuilds a netty request from the zuul `Headers` on every proxied request, and `ClientResponseWriter` rebuilds a netty response on every reply - both via `entries()`, which allocates an `ArrayList` plus a `Header` and a `HeaderName` for every header. `HttpResponseMessageImpl.removeExistingSetCookie` does the same to rebuild a filtered copy.

Now that `Headers` has allocation-free iteration in #2154, this swaps those three over to the new helpers.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientResponseWriter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientResponseWriter.java
@@ -27,7 +27,6 @@ import com.netflix.spectator.api.Registry;
 import com.netflix.zuul.RequestCompleteHandler;
 import com.netflix.zuul.context.CommonContextKeys;
 import com.netflix.zuul.exception.ZuulException;
-import com.netflix.zuul.message.Header;
 import com.netflix.zuul.message.http.HttpRequestInfo;
 import com.netflix.zuul.message.http.HttpRequestMessage;
 import com.netflix.zuul.message.http.HttpResponseMessage;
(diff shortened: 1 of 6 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/ClientResponseWriter.java
```diff
@@ -186,9 +185,7 @@ private HttpResponse buildHttpResponse(HttpResponseMessage zuulResp) {
 
         // Now set all of the response headers - note this is a multi-set in keeping with HTTP semantics
         HttpHeaders nativeHeaders = nativeResponse.headers();
-        for (Header entry : zuulResp.getHeaders().entries()) {
-            nativeHeaders.add(entry.getKey(), entry.getValue());
-        }
+        zuulResp.getHeaders().forEach((name, value) -> nativeHeaders.add(name, value));
```
jguerra, line 188: just confirming - this preserves the header casing?

gavinbunney: Yes correct. The `forEachNormalised` is the non-preserved helper

gavinbunney: Let me add some tests as well here since we should really be validating that contract regardless

jguerra (approved):

PR #2158, opened 2026-06-29
Resolve `FilterConstraint`s once and key the `isConstrained` cache by `Class`

gavinbunney:
`FilterConstraints.isConstrained` runs per filter per message in `BaseZuulFilterRunner.shouldSkipFilter`, so it's a hot map lookup across all constraint-declaring filters. It called `computeIfAbsent(filter.getClass().getName(), lambda)` on every call - a key and a capturing lambda allocated even on a hit, then a second `lookup.get(...)` per declared class to turn each `Class` back into its `FilterConstraint`.

This PR caches the resolved `FilterConstraint[]` keyed by `Class`, so the hot path is a bare `get(filter.getClass())` over pre-resolved instances. 

### Benchmark Results

JMH over the three lookup forms, lower is better.

| Lookup | ns/op | alloc/op |
|---|---|---|
| `getAnnotation().value()` each call | 10.29 | 40 B |
| `computeIfAbsent(getName())` (before) | 4.46 | 24 B |
| `get(Class)` (after) | 2.51 | 0 B |

~1.8x faster and allocation-free on the hot path - the 24 B/op was the per-call capturing lambda.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/filter/FilterConstraints.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/filter/FilterConstraints.java
@@ -19,6 +19,7 @@ package com.netflix.zuul.netty.filter;
 import com.netflix.zuul.FilterConstraint;
 import com.netflix.zuul.filters.ZuulFilter;
 import com.netflix.zuul.message.ZuulMessage;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/filter/FilterConstraints.java
```diff
@@ -36,11 +37,10 @@
 @NullMarked
 public class FilterConstraints {
 
-    @SuppressWarnings("unchecked")
-    private static final Class<? extends FilterConstraint>[] NO_CONSTRAINTS = new Class[0];
+    private static final FilterConstraint[] NONE = new FilterConstraint[0];
 
     private final Map<Class<? extends FilterConstraint>, FilterConstraint> lookup;
-    private final Map<String, Class<? extends FilterConstraint>[]> filterConstraints;
+    private final Map<Class<?>, FilterConstraint[]> filterConstraints;
```
jguerra: With these changes we don't really need to keep FilterConstraint as an array anymore. Want to switch it to an immutable list instead?

gavinbunney: Good call, done

jguerra (approved):

PR #2161, opened 2026-06-30
Switch to event based connection draining

jguerra:
Currently, the connection shutdown logic is tightly coupled to writing responses and requires handlers to check for channel attributes on every write. This PR moves zuul-core to event-based signaling (like our internal WebSocket handling). This new approach gives us better control of shutdowns and gives us flexibility to add new close reasons (e.g. expiration relies on the same eventing as shutdowns)

The current shutdown logic works by setting a promise as a channel attribute. The close handlers then must check for the presence of this flag on every write call, and take some action if it's present. 

The new logic instead relies on firing events down the pipeline (this is internally how we handle websockets). Upon receiving an event, the close logic is largely the same. For http/1.1, a `connection: close` header is added to responses and clients are given some time to close the connection themselves before we close it. For http/2 we instead rely on a two phased go-away process (this is unchanged), however, we now leverage netty's graceful shutdown mechanism to actually close the connection. 

Notable + breaking  changes:
- Http2ConnectionCloseHandler is now a connection level handler - this deals with a lot of the awkwardness that used to exist with it as a stream level handler 
- I removed CLOSE_AFTER_RESPONSE entirely as it relied heavily on the channel attribute approach 
- I introduced the concept of random jitter around connection closing (another concept we use internally) to spread out connection closes during OOS. Jitter is not used for shutdowns.
- Because these closes are now event based they are decoupled from requests. When an event is handled, the graceful shutdown logic triggers immediately instead of waiting for the next response. For http/1.1 this means a connection can be closed immediately if no request is in flight. For http/2, it means the two phase go-away is started immediately upon handling the event. 
- removed the concept of expiration limits during brownout as it was an unneeded complication


zuul-core/src/main/java/com/netflix/netty/common/close/AbstractHttpConnectionExpiryHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+        if (isResponse(msg)) {
+            count++;
+            if (isConnectionExpired(ctx.channel())) {
+                Channel channel = HttpUtils.getMainChannel(ctx);
+                promise.addListener(f -> {
+                    channel.pipeline().fireUserEventTriggered(EXPIRATION_EVENT);
+                });
+            }
```
gavinbunney, line 82: One behavior change here (correct me if I'm wrong?) is that an expired conn won't get the `Connection: Close` header added, rather it would happen on the next request -> triggers expiry -> immediately closed.

zuul-core/src/main/java/com/netflix/netty/common/close/BaseConnectionCloseHandler.java
```diff
@@ -0,0 +1,185 @@
...
+        countStarted(getPort(ctx.channel()));
+        onCloseEvent(ctx, event);
+    }
+
+    private static String closeType(ConnectionCloseEvent event) {
+        return switch (event) {
+            case Graceful ignored -> "GRACEFUL";
+            case GracefulDelayed ignored -> "GRACEFUL_DELAYED";
+        };
+    }
```
gavinbunney: nit: move this into the `Gradeful` and `GradefulDelayed` records themselves as a getter

zuul-core/src/main/java/com/netflix/netty/common/channel/config/CommonChannelConfigKeys.java
gavinbunney, line 76: Unused now?

jguerra: I ended up keeping this for an escape hatch in `Http2ConnectionCloseHandler` after getting paranoid about the old comments of some clients not working with the double go-aways. However, by default it's been enabled for some time so we could probably just remove it entirely

zuul-core/src/main/java/com/netflix/netty/common/close/AbstractHttpConnectionExpiryHandler.java
jguerra: Yeah that is a behavior change. I should be able to restore the old behavior

jguerra: I restored the old behavior

gavinbunney (approved):

PR #2163, opened 2026-07-01
Fix server shutdown hook logic

jguerra:
A constructor cleanup refactoring last year broke jvm shutdown registration logic. Originally if you passed in null it indicated that an external shutdown hook was responsible for the Server lifecycle. However, with the change passing in null results in always registering a shutdown hook. 

 I restored the old behavior and added tests to avoid future breakages

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
@@ -78,6 +78,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/server/Server.java
```diff
@@ -184,32 +187,33 @@ public Server(
...
             EventLoopGroupMetrics eventLoopGroupMetrics,
             EventLoopConfig eventLoopConfig) {
-        this(
-                registry,
-                serverStatusManager,
-                addressesToInitializers,
-                clientConnectionsShutdown,
-                eventLoopGroupMetrics,
-                eventLoopConfig,
-                null);
```
jguerra, line 194: unfortunately I'm unable to pass `new Thread(this::stop, "Zuul-JVM-shutdown-hook");` here because we don't have a `this` reference yet

gavinbunney (approved):

PR #2171, opened 2026-07-03
Replace SessionContext inheritance with composition

jguerra:
SessionContext is-a HashMap<String, Object>. With this PR SessionContext has-a HashMap<String, Object> 

I added a few more maps methods to SessionContext to avoid breakage, but otherwise this is a surprisingly small change


zuul-core/src/test/java/com/netflix/zuul/message/ZuulMessageImplTest.java
```diff
@@ -46,7 +46,7 @@ void testClone() {
 
         assertThat(msg2.getBodyAsText()).isEqualTo(msg1.getBodyAsText());
         assertThat(msg2.getHeaders()).isEqualTo(msg1.getHeaders());
-        assertThat(msg2.getContext()).isEqualTo(msg1.getContext());
+        assertThat(msg2.getContext()).isNotNull();
```
jguerra, line 49: this would require implementing equals on SessionContext. I simplified this check and added a more extensive clone test to SessionContextTest

gavinbunney (approved):
Very nice

PR #2176, opened 2026-07-13
Add new connection pool options for using default buffer sizing

jguerra:
This PR adds a new connection pool configuration that skips configuring TCP buffer sizes on origin connections. When the new flag is true, the connection pool will no longer explicitly set socket options for SNDBUF and RCVBUF, and instead the default OS setting is used. A benefit of relying on the OS is that we can allow it to autotune the buffer sizes for us

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ConnectionPoolConfig.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ConnectionPoolConfig.java
@@ -23,26 +23,45 @@ import com.netflix.zuul.origins.OriginName;
  */
 public interface ConnectionPoolConfig {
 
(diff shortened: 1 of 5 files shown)


zuul-core/src/main/java/com/netflix/zuul/netty/connectionpool/ConnectionPoolConfig.java
```diff
@@ -23,27 +23,46 @@
  */
 public interface ConnectionPoolConfig {
 
-    /* Origin name from connection pool */
+    /**
+     * Origin name from connection pool.
```
AlexanderEllis, line 27: Controversial comment formatting!

AlexanderEllis (approved):
LGTM!

PR #2186, opened 2026-07-17
Strip X-Forwarded-Host, X-Forwarded-Prefix and Forwarded from untrusted connections

### LINKED ISSUE #2170: [Security] `StripUntrustedProxyHeadersHandler` does not strip `X-Forwarded-Host` header
The `StripUntrustedProxyHeadersHandler` class is documented to "strip out any X-Forwarded-* headers from inbound http requests if connection is not trusted" (class Javadoc, line 37). However, the actual implementation omits several critical proxy headers from the strip set, most notably `X-Forwarded-Host`, which is then trusted and forwarded to origin services.

This creates a gap between the documented security posture and actual behavior when operators configure `AllowWhen.NEVER` (the recommended secure configuration per the Zuul wiki for deployments without a trusted ELB in front).

## Affected Code

**File:** `zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java`

**Lines 50-55:**
```java
private static final Collection<AsciiString> HEADERS_TO_STRIP = Sets.newHashSet(
        new AsciiString("x-forwarded-for"),
        new AsciiString("x-forwarded-port"),
        new AsciiString("x-forwarded-proto"),
        new AsciiString("x-forwarded-proto-version"),
        new AsciiString("x-real-ip"));
```

**Missing from strip set:**
- `x-forwarded-host`
- `x-forwarded-prefix`
- RFC 7239 `forwarded`

## Impact

When `StripUntrustedProxyHeadersHandler` is configured with `AllowWhen.NEVER` (the recommended secure configuration per the Zuul wiki), operators reasonably expect that client-supplied `X-Forwarded-*` headers are removed. However, `X-Forwarded-Host` is not stripped.

**Note:** The framework default is `AllowWhen.ALWAYS` per `CommonChannelConfigKeys.java`, but the security-conscious configuration recommended by Netflix is `NEVER` for deployments without a trusted load balancer.

The method `HttpRequestMessageImpl.getOriginalHost()` (lines 505-515) then trusts `X-Forwarded-Host` with top priority over the actual `Host` header:

```java
static String getOriginalHost(Headers headers, String serverName) throws URISyntaxException {
    String xForwardedHost = headers.getFirst(HttpHeaderNames.X_FORWARDED_HOST);
    if (xForwardedHost != null) {
        return xForwardedHost;  // client-controlled, returned immediately
    }
    // ... fallback to Host header
}
```

This value is then:
1. Forwarded to origin services via `ProxyUtils.addXForwardedHeaders()` (with `overwrite=false` preserving the client value)
2. Used in `reconstructURI()` (line 645) where it appears in access logs and response headers (for example, in the sample's `X-Originating-URL`)

### Exploitation Scenario

Consider a deployment configured with `AllowWhen.NEVER` (following Netflix's security recommendation) to reject client proxy headers:

**Attacker request:**
```http
GET /account/reset HTTP/1.1
Host: gateway.example.com
X-Forwarded-Host: evil.attacker.com
```

**What happens:**
- The `stripXFFHeaders()` method removes `X-Forwarded-For`, `-Port`, `-Proto`, and `-Real-IP`
- But `X-Forwarded-Host: evil.attacker.com` remains untouched
- The `getOriginalHost()` method returns `evil.attacker.com`
- The origin service receives: `X-Forwarded-Host: evil.attacker.com`

If the origin service builds absolute URLs from `X-Forwarded-Host` (password reset links, redirects) or uses it as a cache key, this enables:
- Password reset link poisoning
- Web cache poisoning (if cache is keyed on host)
- Open redirect (if origin trusts the header for redirects)

## Scope Note

This finding is about defense-in-depth and hardening. The actual impact depends on downstream origin behavior:
- Zuul itself is not directly vulnerable (routing is VIP-based, not header-based)
- Origin services that trust `X-Forwarded-Host` for URL construction or caching are at risk
- Operators deploying Zuul with the `NEVER` policy reasonably expect full `X-Forwarded-*` protection based on the Javadoc

## Recommended Fix

**Option 1 (minimal change):** Add the missing headers to `HEADERS_TO_STRIP`:

```java
private static final Collection<AsciiString> HEADERS_TO_STRIP = Sets.newHashSet(
        new AsciiString("x-forwarded-for"),
        new AsciiString("x-forwarded-host"),        // ADD THIS
        new AsciiString("x-forwarded-prefix"),      // ADD THIS
        new AsciiString("x-forwarded-port"),
        new AsciiString("x-forwarded-proto"),
        new AsciiString("x-forwarded-proto-version"),
        new AsciiString("x-real-ip"),
        new AsciiString("forwarded"));              // ADD THIS (RFC 7239)
```

**Option 2 (defense-in-depth):** Make `getOriginalHost()` prefer the validated `Host` header when `AllowWhen.NEVER` is active, rather than trusting `X-Forwarded-Host` unconditionally.

**Option 3 (documentation):** If the current behavior is intentional, update the class Javadoc to clarify that `X-Forwarded-Host` and `Forwarded` are not stripped even in `NEVER` mode.

## Verified Safe Areas (for context)

I reviewed the following areas and confirmed they are not vulnerable:
- HTTP/1.1 request smuggling protections (`Http1FramingEnforcingHandler`) are solid
- HTTP/2 per-stream content-length enforcement is working correctly
- Multiple `Host` headers are properly rejected
- Routing is VIP-based (no SSRF via `Host` or `X-Forwarded-Host` injection)

## Steps to Reproduce

1. Deploy Zuul with `StripUntrustedProxyHeadersHandler(AllowWhen.NEVER)`
2. Send the following request:
   ```http
   GET / HTTP/1.1
   Host: gateway.example.com
   X-Forwarded-Host: attacker.com
   ```
3. Observe in origin logs or Zuul's `X-Originating-URL` response header that it shows `http://attacker.com/` (the attacker-controlled host)

## References

- RFC 7230 § 5.4: Host header handling
- RFC 7239: Forwarded HTTP Extension (standardized proxy headers)
- OWASP: Web Cache Poisoning via Host Header Injection

## Additional Context

This gap was discovered during a security audit of Zuul's proxy header handling as part of research into API gateway hardening patterns. The audit also confirmed that Zuul's HTTP/1.1 and HTTP/2 request smuggling protections are working correctly (see "Verified Safe Areas" above).

## Disclosure

- **Discovered:** 2026-07-03
- **Researcher:** refleeexzz
- **Coordination:** Reporting publicly via GitHub as this is a defense-in-depth gap with no immediate production impact to Netflix deployments

I'm happy to provide a PR implementing any of the recommended fixes if the maintainers would like. Thank you for maintaining Zuul and for considering this report!

fool1280:
## Summary

Fixes #2170.

`StripUntrustedProxyHeadersHandler` is documented to "Strip out any X-Forwarded-* headers from inbound http requests if connection is not trusted", but its `HEADERS_TO_STRIP` set omitted three headers:

- `X-Forwarded-Host`
- `X-Forwarded-Prefix`
- `Forwarded` (RFC 7239)

`X-Forwarded-Host` is the impactful one: `HttpRequestMessageImpl.getOriginalHost()` prefers it over the validated `Host` header and it is propagated to origins. Since it was never stripped (even under the security-hardened `AllowWhen.NEVER` default), an untrusted client could spoof it, enabling host-header-based attacks in downstream origins that build URLs or cache keys from the value (web cache poisoning, password-reset link poisoning, open redirect).

Zuul itself is not directly vulnerable (routing is VIP-based); this is a defense-in-depth hardening that aligns the handler's behavior with its documented contract. `X-Forwarded-Prefix` and `Forwarded` are added for completeness even though they are not currently consumed internally.

## Change

Add `x-forwarded-host`, `x-forwarded-prefix`, and `forwarded` to `HEADERS_TO_STRIP`.

## Testing

- Added `strip_hostRewritingHeaders` to `StripUntrustedProxyHeadersHandlerTest` asserting the three headers are removed by `stripXFFHeaders`.
- `./gradlew :zuul-core:test --tests "*StripUntrustedProxyHeadersHandlerTest"` passes.
- `./gradlew :zuul-core:spotlessCheck` passes.

### DIFF
--- a/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
@@ -49,10 +49,13 @@ public class StripUntrustedProxyHeadersHandler extends ChannelInboundHandlerAdap
 
     private static final Collection<AsciiString> HEADERS_TO_STRIP = Sets.newHashSet(
             new AsciiString("x-forwarded-for"),
+            new AsciiString("x-forwarded-host"),
             new AsciiString("x-forwarded-port"),
+            new AsciiString("x-forwarded-prefix"),
             new AsciiString("x-forwarded-proto"),
             new AsciiString("x-forwarded-proto-version"),
-            new AsciiString("x-real-ip"));
+            new AsciiString("x-real-ip"),
+            new AsciiString("forwarded"));
 
     private final AllowWhen allowWhen;
 
--- a/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
+++ b/zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
@@ -17,7 +17,6 @@
 package com.netflix.netty.common.proxyprotocol;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Sets;
 import com.netflix.config.DynamicStringListProperty;
 import com.netflix.netty.common.ssl.SslHandshakeInfo;
 import com.netflix.zuul.netty.server.ssl.SslHandshakeInfoHandler;
@@ -32,6 +31,7 @@ import io.netty.handler.ssl.ClientAuth;
 import io.netty.util.AsciiString;
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Strip out any X-Forwarded-* headers from inbound http requests if connection is not trusted.
@@ -47,7 +47,7 @@ public class StripUntrustedProxyHeadersHandler extends ChannelInboundHandlerAdap
         NEVER
     }
 
-    private static final Collection<AsciiString> HEADERS_TO_STRIP = Sets.newHashSet(
+    private static final Collection<AsciiString> HEADERS_TO_STRIP = Set.of(
             new AsciiString("x-forwarded-for"),
             new AsciiString("x-forwarded-host"),
             new AsciiString("x-forwarded-port"),
--- a/zuul-core/src/test/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandlerTest.java
+++ b/zuul-core/src/test/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandlerTest.java
@@ -157,6 +157,20 @@ class StripUntrustedProxyHeadersHandlerTest {
         assertThat(headers.contains("x-forwarded-for")).isFalse();
     }
 
+    @Test
+    void strip_hostRewritingHeaders() {
+        StripUntrustedProxyHeadersHandler stripHandler = getHandler(AllowWhen.MUTUAL_SSL_AUTH);
+
+        headers.add("x-forwarded-host", "evil.attacker.com");
+        headers.add("x-forwarded-prefix", "/evil");
+        headers.add("forwarded", "host=evil.attacker.com");
+        stripHandler.stripXFFHeaders(msg);
+
+        assertThat(headers.contains("x-forwarded-host")).isFalse();
+        assertThat(headers.contains("x-forwarded-prefix")).isFalse();
+        assertThat(headers.contains("forwarded")).isFalse();
+    }
+
     private StripUntrustedProxyHeadersHandler getHandler(AllowWhen allowWhen) {
         return spy(new StripUntrustedProxyHeadersHandler(allowWhen));
     }


zuul-core/src/main/java/com/netflix/netty/common/proxyprotocol/StripUntrustedProxyHeadersHandler.java
```diff
@@ -49,10 +49,13 @@ public enum AllowWhen {
 
     private static final Collection<AsciiString> HEADERS_TO_STRIP = Sets.newHashSet(
```
jguerra: while you're here - can you switch this to an immutable Set.of call instead?

fool1280: Done!

jguerra (approved):

PR #2206, opened 2026-08-11
Switch ConnCounter to rely on PolledMeters

jguerra:
`ConnCounter` tracks connection counts by manually updating a gauge. This approach can lead to negative connection counts when gauges expire. The gauges used in this class expire 15 mins after the last update. Stable http/2 connections to origins are especially susceptible to getting lost.

The new approach relies on a PolledMeter monitoring an AtomicLong. Each event loop independently tracks its own counts (internally gauges maintain a list of all monitored numbers with the same id, and sum them up). I did this to simplify the cleanup logic

### DIFF
--- a/zuul-core/src/main/java/com/netflix/zuul/monitoring/ConnCounter.java
+++ b/zuul-core/src/main/java/com/netflix/zuul/monitoring/ConnCounter.java
@@ -46,6 +46,10 @@ public final class ConnCounter {
 
     private static final Attrs EMPTY = Attrs.newInstance();
 
+    /**
+     * PER_EVENT_LOOP_COUNTERS exists to reduce the number of PolledMeters that are created. Any given Id will have a
+     * PolledMeter created for every event loop, with spectator responsible for summing up the values.
+     */
     private static final ThreadLocal<Map<Id, AtomicInteger>> PER_EVENT_LOOP_COUNTERS =
             ThreadLocal.withInitial(HashMap::new);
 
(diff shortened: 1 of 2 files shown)


zuul-core/src/main/java/com/netflix/zuul/monitoring/ConnCounter.java
```diff
@@ -110,45 +103,45 @@ public void increment(String event, Attrs extraDimensions) {
...
     }
 
     public void decrement(String event) {
         Objects.requireNonNull(event);
-        Gauge gauge = counts.remove(event);
-        if (gauge == null) {
-            // TODO(carl-mastrangelo): make this throw IllegalStateException after verifying this doesn't happen.
+        Id id = eventToIdLookup.remove(event);
+
+        if (id == null) {
```
gavinbunney, line 133: From a state machine perspective, does this mean if you do two increments, after the first decrement you won't be able to decrement again after that? Wondering the value of it being a `number` as it seems more like it operates as a bool unless I'm missing some part of the internal usage of the counter?

jguerra: A `ConnCounter` is per channel so increment/decrement should truly only be called once per-event type, e.g. "active". This check is really just a guard for a programming error - I could switch it to throw like the old TODO says. `eventToIdLookup` is needed because we lose the actual `Id` after increment, and need to be able to drive things based on just the eventName.

gavinbunney (approved):

PR #2213, opened 2026-09-06
Match query param names verbatim in HttpQueryParams.get

lenamonj:
`HttpQueryParams.get(String)` lower-cased the lookup key while names are stored verbatim, so every parameter whose name carries an uppercase letter returned an empty list, and `get` disagreed with `getFirst` and `contains`, which match verbatim. `containsIgnoreCase` folded only the argument, so it missed a stored name with any uppercase letter.

`get` now matches the stored name, and `containsIgnoreCase` compares against the stored names case-insensitively. Three tests cover both; the `get` cases fail on master.


jguerra (approved):
Thanks for the contribution!

PR #2215, opened 2026-09-11
Validate zstd compression

sullis:
Adds an integration test that validates zstd response compression end-to-end.

`zstdOnly()` sends `Accept-Encoding: zstd`, asserts Zuul responds with `Content-Encoding: zstd`, and decompresses the body with `com.github.luben.zstd.Zstd` to confirm the payload round-trips correctly. This mirrors the existing brotli-only test, which only checked headers/length without verifying the compressed bytes are actually valid.

- Adds `com.github.luben:zstd-jni` as a `testImplementation` dependency in `zuul-integration-test` (version comes from the `netty-parent` BOM already on the test classpath).
- `Zstd.ensureAvailability()` guards the test so it fails loudly rather than silently if native support is missing.

The `dependencies.lock` churn (jackson 2.21.5 → 2.21.6, spectator 1.10.4 → 1.10.7) is from regenerating the lock files.

### DIFF
--- a/zuul-core/dependencies.lock
+++ b/zuul-core/dependencies.lock
@@ -12,10 +12,10 @@
     },
     "compileClasspath": {
         "com.fasterxml.jackson.core:jackson-core": {
-            "locked": "2.21.5"
+            "locked": "2.21.6"
         },
         "com.fasterxml.jackson.core:jackson-databind": {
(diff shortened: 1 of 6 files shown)


jguerra (approved):
Thanks Sean!

PR #2221, opened 2026-09-18
Avoid per-header allocation in zuul-to-netty handoff

gavinbunney:
Zuul currently copies origin requests and client responses into Netty's general-purpose `DefaultHttpHeaders`, allocating an entry object for every header immediately before encoding.

This PR adds `ZuulToNettyHttpHeaders`, a mutable sequential representation scoped to these two copy-and-encode boundaries. It preserves required mutation semantics and duplicate ordering, with H1 byte-equivalence and H2 conversion-equivalence tests.

#### Benchmark results

The isolated handoff is the behavior changed by this PR:

| Profile | Current B/op | New B/op | Allocation saved | Current ops/s | New ops/s | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| Request - 80 headers, 16 KB | 3,384 | 680 | **80%** | 893K | 6.18M | **6.92x** |
| Response - 28 headers, 1.8 KB | 1,320 | 280 | **79%** | 1.78M | 17.52M | **9.87x** |

End-to-end results include destination construction plus Netty encoding or conversion:

| Profile | Path | Current B/op | New B/op | Allocation saved | Current ops/s | New ops/s | Speedup |
|---|---|---:|---:|---:|---:|---:|---:|
| Request - 80 headers, 16 KB | H1 encoding | 3,652 | 932 | **74%** | 234K | 306K | **1.30x** |
| Request - 80 headers, 16 KB | H2 header conversion | 14,272 | 11,536 | **19%** | 237K | 243K | **1.02x** |
| Response - 28 headers, 1.8 KB | H1 encoding | 1,588 | 524 | **67%** | 821K | 1.05M | **1.27x** |
| Response - 28 headers, 1.8 KB | H2 header conversion | 4,616 | 3,552 | **23%** | 942K | 1.36M | **1.44x** |

H2 measures Netty header conversion rather than HPACK or wire encoding.


zuul-core/src/main/java/com/netflix/zuul/netty/ZuulToNettyHttpHeaders.java
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    @Override
+    public int size() {
+        return this.size;
+    }
+
+    @Override
+    public Set<String> names() {
+        Set<String> names = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
```
jguerra: the sort seems kind of sneaky. Maybe a LinkedHashSet would be more appropriate?

gavinbunney: Nice pickup - updated to use a `LinkedHashSet` (`DefaultHttpHeaders` uses one as well so it is more consistent)

jguerra (approved):

